PR #44, opened 2019-01-24
OHARA-1317 Fix connector not updating graph issue

konekoya:
- update graph name when an update is happened in connector


konekoya:
retry manager

konekoya:
retry manager

konekoya:
Thanks for reviewing @eechih :)

PR #46, opened 2019-01-24
OHARA-1397 Add a column to display worker cluster that the pipeline is in use in pipeline list page

konekoya:
- Added a new column to display cluster name
- Removed index column since it's useless now
- Fixed related unit tests


konekoya:
retry manager

konekoya:
Fail build is not related

PR #58, opened 2019-01-25
OHARA-1415 Add connector failed message to Pipeline APIs

chia7712:


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/ConnectorRoute.scala
```diff
@@ -145,8 +145,8 @@ private[configurator] object ConnectorRoute extends SprayJsonSupport {
                       if (!topicInfos.exists(_.id == t))
                         throw new IllegalArgumentException(
                           s"$t is invalid. actual:${topicInfos.map(_.id).mkString(",")}"))
+                    if (connectorConfig.topics.isEmpty) throw new IllegalArgumentException("topics is required")
```
konekoya: are :)

PR #68, opened 2019-01-26
OHARA-1403 Remove all connectors connected to a deleted pipeline

chia7712:
two new rules is added to DELETE Pipeline API.

1. If any object in pipeline has state (it means the object exists in somewhere), it is illegal to delete pipeline.
1. If all objects have no state, deleting pipeline will delete all objects(exclude topic and streamapp) also


konekoya:
Awesome, thanks for the help @chia7712

PR #77, opened 2019-01-28
OHARA-1473 Add tooltips for control buttons and toolbar items

konekoya:
- Add `react-tooltip` package
- Add tooltips for toolbar items and some control buttons


konekoya:
retry manager

PR #81, opened 2019-01-28
OHARA-1441 Instead of throwing exception, we should return empty list…

chia7712:
… when failing to fetch connectors from a starting worker cluster


chia7712:
retry it

PR #87, opened 2019-01-29
OHARA-1387 Delete individual connector from pipeline

eechih:
可以單獨刪除 FtpSource, JdbcSource, FtpSink, HdfsSink, StreamApp
刪除成功後，自動更新 Pipeline

PR #88, opened 2019-01-29
OHARA-1447 Fix tab component react warning and remove related styles

konekoya:
- Fix warning
- Remove unused styles


konekoya:
@eechih I'm seeing some react related warning stating that the prop `showShadow` is not valid DOM props. Not entirely sure about why this is happening but since we don't need these style anymore (at least not for now), I'm removing its related styles and props. Let me know if you have any other concerns :-)

konekoya:
I think you're including the fix in #87, right? if that's true. I'm going to close this PR then :)

eechih:
> I think you're including the fix in #87, right? if that's true. I'm going to close this PR then :)

Yes.

PR #90, opened 2019-01-29
OHARA-1449 Limit the port range for worker service, from 5000 to 65535

eechih:
新增 Worker Cluster 時，Port 應當限制在 5000 ~ 65535 之間


ohara-manager/client/src/components/pages/Services/WorkerNewModal/WorkerNewModal.js
```diff
@@ -51,6 +51,13 @@ class WorkerNewModal extends React.Component {
   };
 
   onSubmit = async (values, form) => {
+    const { clientPort } = values;
+    if (clientPort < 5000 || clientPort > 65535) {
+      toastr.error(
+        'Invalid port given. The port has to be a value from 5000 to 65535.',
```
konekoya: Could we remove the `given`, so it would be 

`Invalid port! The port has to be a value between 5000 and 65535.`

chia7712:
> Could take a look at the port range and see if that's what you're expecting

> 5000 is ok to me

konekoya:
Thanks for the PR @eechih

PR #91, opened 2019-01-29
OHARA-1451 NPE happens on receiving row data from connector

chia7712:


chia7712:
retry connector

chia7712:
retry it

PR #92, opened 2019-01-29
OHARA-1455 Extend pipeline API so it can accept an array of connector…

chia7712:
… ids instead of a single one


konekoya:
Thanks for the quick fix @chia7712 I'm going to work on this in tomorrow morning. Will let you know if there's anything else that I need help from you.

chia7712:
> Thanks for the quick fix @chia7712 I'm going to work on this in tomorrow morning. Will let you know if there's anything else that I need help from you.

update front-end ASAP :)

konekoya:
Yep, absolutely. 

/cc @vitojeng after this PR. The pipeline graph would probably break, therefore affecting demo env as well. I'll need to make some changes in [OHARA-1385](https://is-land.atlassian.net/browse/OHARA-1385). Sorry about the trouble!

PR #117, opened 2019-02-01
OHARA-1480 add variable to tag of build.yml (docker-compose file)

chia7712:
After merging this PR, delivery-ohara can generate different tags for each ohara's image. Also, deploy-ohara is able to start services by "specified" tag.


vitojeng:
Conflict should be due to #113 just merged

vitojeng:
overall pretty good.
I'm +1 after conflict solved.

PR #119, opened 2019-02-01
OHARA-1465 Support delete function in pipeline topic graph

eechih:
允許使用者刪除沒有任何 connection 的 topic


ohara-manager/client/src/components/pages/PipelinePage/connectors/Topic.js
```diff
@@ -16,21 +16,17 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import { get, isEmpty, some, includes, omit } from 'lodash';
```
konekoya: Hmm. I've been thinking about this lately. Let's remove all the lodash utility functions from `commonUtils` and use lodash directly in components that are needed these utils. What do you think about this?

ohara-manager/client/src/components/pages/PipelinePage/connectors/Controller/Controller.js
```diff
@@ -17,16 +17,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactTooltip from 'react-tooltip';
+import { noop, includes } from 'lodash';
```
konekoya: Not super familiar with noop. Would you mind elaborate a bit why use this over primitive values like `String`, `Number` or `Null`? Thanks :)

ohara-manager/client/src/components/pages/PipelinePage/connectors/Topic.js
konekoya: Let's talk about this in person...

konekoya: Alright, that's make this happend :)

ohara-manager/client/src/components/pages/PipelinePage/connectors/Controller/Controller.js
eechih: > Not super familiar with `noop`. Would you mind elaborate a bit why use this over primitive values like `String`, `Number` or `Null`? Thanks :)

const noop = () => undefined;

ohara-manager/client/src/components/pages/PipelinePage/connectors/Topic.js
eechih: 👍

ohara-manager/client/src/components/pages/PipelinePage/connectors/Controller/Controller.js
konekoya: Pretty cool. Maybe I'll need this in the future. Thanks for explaining all that.

konekoya:
Thanks :)

PR #120, opened 2019-02-01
OHARA-1482 The new cluster button in services page should be disabled on page load

konekoya:
- Fix the bug
- Add tests


konekoya:
retry manager

PR #129, opened 2019-02-08
OHARA-689 Use styled-component theme provider to simplify the CSS color variables sharing 

konekoya:


chia7712:
run

konekoya:
By using ThemePovider, we no longer need to import css variables in each component :)

/cc @eechih

PR #144, opened 2019-02-12
[WIP]OHARA-595 Metadata of RowConnector

g1geordie:
here has "a lot of mess" .  please view and give some suggestion

what I want to do  is
The Connector will bind our supportype data from front end
and then pass the object to task

We use this binding object to extract metadata and give front end info


build.gradle
```diff
@@ -129,7 +129,7 @@ subprojects {
       tests.extendsFrom archives
     }
     task testJar(type: Jar, dependsOn: testClasses) {
-      archiveClassifier.set('tests')
+//      archiveClassifier.set('tests')
```
chia7712: Why? Are you using gradle 4.x ? If so, please upgrade to 5.x

ohara-configurator/src/main/scala/com/island/ohara/configurator/jar/JarStore.scala
```diff
@@ -72,6 +72,8 @@ trait JarStore extends Releasable {
 
   def jarInfos(ids: String): Future[Seq[JarInfo]] = jarInfos().map(_.filter(p => ids.contains(p.id)))
 
+  def jarInfos(ids: Seq[String]): Future[Seq[JarInfo]] = jarInfos().map(_.filter(p => ids.contains(p.id)))
```
chia7712: Nice one. Seems the variant (jarInfos(ids: String)) is a wrong method. Please remove it

build.gradle
g1geordie: Yes ,you r right . I have upgrade it

ohara-kafka/src/main/java/com/island/ohara/kafka/connector/TaskConfig2.java
```diff
@@ -0,0 +1,128 @@
...
+import com.island.ohara.common.annotations.Optional;
+import com.island.ohara.common.data.Column;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/** this class carries all required configs for row connectors. */
+public class TaskConfig2<T> {
```
chia7712: this name is not acceptable...

chia7712: Please describe the usage of this class...

chia7712:
We need more discussion. see https://github.com/oharastream/ohara/issues/297

PR #145, opened 2019-02-12
OHARA-1510 Update the docker version to 18.09

chia7712:


konekoya:
@chia7712 does this mean that we need to upgrade our docker as well :)

chia7712:
> does this mean that we need to upgrade our docker as well :)

yep

PR #149, opened 2019-02-12
OHARA-1513 JDBC connector is missing a handler in its property form

konekoya:


ohara-manager/client/src/components/pages/PipelinePage/Connectors/JdbcSource.js
```diff
@@ -557,7 +557,7 @@ class JdbcSource extends React.Component {
                 selected={currWriteTopic}
                 width="100%"
                 data-testid="write-topic-select"
-                handleChange={this.handleChangeSelect}
+                handleChange={this.handleSelectChange}
```
chia7712: 這兩個命名好銷魂 有啥差別嗎

konekoya: AFAIK, it's just a naming convention of the React community... but I always get confused with them 🤦‍♂️

konekoya:
Thanks for reviewing @chia7712 :)

PR #153, opened 2019-02-12
OHARA-1515 Disable user to create multi broker clusters on same zooke…

chia7712:
…eper cluster


chia7712:
retry it

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/BrokerRoute.scala
```diff
@@ -17,47 +17,58 @@
...
+                  .filter(_.isInstanceOf[ZookeeperClusterInfo])
+                  .find(_.name == zkName)
+                  .map(_.name)
+                  .getOrElse(throw new NoSuchElementException(s"$zkName doesn't exist"))
+              }
+              .getOrElse {
+                val zkClusters = clusters.keys.filter(_.isInstanceOf[ZookeeperClusterInfo])
+                if (zkClusters.size != 1)
+                  throw new IllegalArgumentException(
+                    s"You don't specify the zk cluster for ${req.name}, and there is no default zk cluster")
```
konekoya: I think it should be *didn't*

chia7712:
failed test is unrelated to this PR. will address @konekoya 's comment and then merge it.

PR #158, opened 2019-02-13
OHARA-1519 Enable to remove connector/pipeline from deleted wk cluster

chia7712:


chia7712:
retry it

chia7712:
retry

ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestConnectorsRoute.scala
```diff
@@ -159,48 +156,66 @@ class TestConnectorsRoute extends SmallTest with Matchers {
...
+    val connector = result(
+      connectorApi.add(
+        ConnectorCreationRequest(name = methodName,
+                                 className = "jdbc",
+                                 schema = Seq.empty,
+                                 configs = Map("c0" -> "v0", "c1" -> "v1", "c2" -> "v2"),
+                                 topics = Seq.empty,
+                                 numberOfTasks = 1)))
+
+    val wk = result(configurator.clusterCollie.workerCollie().remove(connector.workerClusterName))
```
saivirtue: just curious, what happened if we add the worker cluster again?

chia7712: > just curious, what happened if we add the worker cluster again?

It will fail if we added another wk cluster with same name.

chia7712:
retry it

ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestConnectorsRoute.scala
saivirtue: so the following scenario will throw exception ?
1. add connector ct1 with wk cluster wk1
2. stop wk1
3. remove ct1
4. start wk1 again

chia7712: If you "have" removed the wk (cluster), it is legal to add wk (cluster) again. (there is no identical name, right?)

chia7712:
retry configurator

chia7712:
"TestK8SCollie.testClusters" is unrelated to this pr

PR #172, opened 2019-02-15
OHARA-1530 StreamApp Property API could Remove Request Validation

saivirtue:
將StreamApp參數檢查移動到Start 階段再做檢查


konekoya:
Thanks for the quick fix @saivirtue!

saivirtue:
retry connector

PR #178, opened 2019-02-15
OHARA-1512 Start and stop stream app connector with real stream app APIs

konekoya:
- Use the real stream app APIs
- Remove delete button in stream app property form as we won't ship the feature in 0.2

**Please note that due to stream app APIs are still under development. Starting and stopping a stream app is always failing right now.**


konekoya:
No worries @saivirtue :-)

konekoya:
retry manager

PR #180, opened 2019-02-15
OHARA-1532 Start pipeline button should also include stream apps

konekoya:
- Fix pipeline start and stop actions
- Add more tests


konekoya:
retry manager

PR #221, opened 2019-02-19
OHARA-1549 Fix validate node api in fake mode

chia7712:


konekoya:
Thanks for the quick fix @chia7712! 

/cc @eechih

PR #278, opened 2019-02-20
OHARA-1561 Prevent the same topic from adding into the pipeline graph…

konekoya:
… again


konekoya:
@vitojeng Adding some new logic to prevent topic from adding into the graph more than once. Thanks for raising this up!

konekoya:
Thanks for reviewing @vitojeng!

PR #282, opened 2019-02-20
OHARA-1567 Ohara StreamApp cannot start normally

saivirtue:


docker/streamapp.dockerfile
```diff
@@ -71,6 +71,4 @@ WORKDIR /opt/ohara
 
 COPY --from=deps /opt/lib/* /opt/ohara/
 
-VOLUME ["/opt/ohara/streamapp"]
-
-ENTRYPOINT ["/tini", "--", "/usr/bin/java", "-cp", "/opt/ohara/*:/opt/ohara/streamapp/*"]
\ No newline at end of file
+ENTRYPOINT ["/tini", "--", "/usr/bin/java", "-cp", "/opt/ohara/*"]
```
chia7712: 所有streamapp環境下的libs都放在/opt/ohara/底下？

ohara-streams/src/main/java/com/island/ohara/streams/StreamApp.java
```diff
@@ -127,31 +131,35 @@ public void stop() throws Exception {}
...
 
-    File dir = new File(jarPath);
-    if (!dir.isDirectory() || !dir.exists())
-      throw new IllegalArgumentException("the parameter should be a valid directory");
-    File[] files = dir.listFiles((File d, String name) -> name.endsWith(".jar"));
-    if (files == null || files.length == 0) {
-      throw new RuntimeException("the path is not contained any jars");
+    URL url = new URL(jarUrl);
+    try (InputStream stream = url.openStream()) {
+      ReadableByteChannel channel = Channels.newChannel(stream);
```
chia7712: Consider reusing code. https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#copyInputStreamToFile(java.io.InputStream,%20java.io.File)

docker/streamapp.dockerfile
saivirtue: 對，包含需要的libs

PR #286, opened 2019-02-21
OHARA-1575 Replace the favicon with the new ohara logo

konekoya:


konekoya:
Seems like GitHub doesn't support `.ico` file preview...

PR #309, opened 2019-02-21
OHARA-1593 ClusterCollieImpl should add --network=host to worker cluster by default

chia7712:


ohara-agent/src/main/scala/com/island/ohara/agent/ClusterCollieImpl.scala
```diff
@@ -647,6 +649,13 @@ private object ClusterCollieImpl {
                     ))
                     .name(hostname)
                     .route(route ++ existRoute)
+                    // we use --network=host for worker cluster since the connectors run on worker cluster may need to
+                    // access external system to request data. In ssh mode, dns service "may" be not deployed.
+                    // (BTW, I hate the guys who run our powerful ohara on a fucking non-production env. by chia)
```
saivirtue: hmm...since this is an "open source", I think may be we can use more mild statement?

chia7712: > hmm...since this is an "open source", I think may be we can use more mild statement?

you are right :)

chia7712:
retry agent

PR #314, opened 2019-02-22
OHARA-1596 Edit message of StreamApp api to be more clarified

saivirtue:


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -189,9 +189,12 @@ private[configurator] object StreamRoute {
           } ~
             path(Segment) { id =>
               //add property is impossible, we need streamApp id first
-              post { complete("unsupported method") } ~
+              post { complete("You should upload a jar first and use PUT method to update properties.") } ~
```
chia7712: Should we use explicit error code? How ablout 400 Bad Request?

chia7712:
IIRC, there are other similar issues in StreamRoute. Could you fix them in this PR?

saivirtue:
> IIRC, there are other similar issues in StreamRoute. Could you fix them in this PR?

sure.

PR #315, opened 2019-02-22
OHARA-1581 Steam app property form should update its topic list when the list updates

konekoya:


konekoya:
Thanks for reviewing @saivirtue!

PR #310, opened 2019-02-22
OHARA-1201 Substitute connectors API for sources API and sinks API

chia7712:

PR #325, opened 2019-02-25
OHARA-1606 Remove useless "sources" and "sinks" APIs

chia7712:
this is a follow-up of #310


chia7712:
當笨蛋了...不小心提交到無關的程式碼...

PR #330, opened 2019-02-25
OHARA-1616 Validation APIs should use request body to replace query p…

chia7712:
…arameter


chia7712:
@eechih Please take a look

PR #338, opened 2019-02-27
OHARA-1627 Configurator should check docker on remote node when handl…

chia7712:
…ing validation request


chia7712:
retry agent

PR #339, opened 2019-02-27
OHARA-1624 [Configurator] On FTP connector, empty properties should b…

saivirtue:
…e handled


saivirtue:
retry

chia7712:
The new tests are great. However, could you add UT for toMap() only.

saivirtue:
> The new tests are great. However, could you add UT for toMap() only.

do you mean the property test part?

chia7712:
> do you mean the property test part?

yep

saivirtue:
done.

PR #341, opened 2019-02-27
OHARA-1623 On FTP connector, drop down list default value should be handled

eechih:
Set default values for connector when file_encoding, task field is empty


konekoya:
Thanks for the fix @eechih :)

PR #350, opened 2019-03-04
OHARA-1639 Fail to start backend due to lack of mysql libraries

chia7712:


konekoya:
Thanks for the quick fix :)

chia7712:
test image on local. Works well.

PR #358, opened 2019-03-05
OHARA-1651 Add UCP requires IPv4 IP Forwarding for docker into the README.md

eechih:


docker/README.md
```diff
@@ -74,4 +74,12 @@ We are talking about tests, right? We don't care about how to shutdown services
...
 ```
 We are talking about tests, right? You should have a machine for testing only so it is ok to remove all containers quickly.
-That does simplify your work and life.
\ No newline at end of file
+That does simplify your work and life.
+
+## How to enable IPv4 IP Forwarding
+
+```
+$ sudo sysctl -w net.ipv4.conf.all.forwarding=1
```
chia7712: 麻煩註明一下這個方式所改的設定並不是permanent

eechih: 設定方式改成 permanent

PR #360, opened 2019-03-05
OHARA-1650 Start pipeline action should return correct messages

konekoya:
- Fix the bug
- Remove an outdated comment and test


konekoya:
retry manager

konekoya:
retry manager

PR #364, opened 2019-03-05
OHARA-1661 Configurator should expose bound hostname and port by default

chia7712:


chia7712:
retry it

chia7712:
unrelated error.

chia7712:
@saivirtue Thanks for reviews!

PR #365, opened 2019-03-05
OHARA-1656 Fix pipeline status is not correctly displaying bug

konekoya:
- Fix the status bug
- Add tests


konekoya:
retry manager

PR #374, opened 2019-03-06
OHARA-1667 StreamApp could not resolve hostname

saivirtue:


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -361,6 +361,11 @@ private[configurator] object StreamRoute {
                                               StreamApi.TO_TOPIC_KEY -> data.toTopics.head
                                             )
                                           )
+                                          .route(
+                                            Map(
```
chia7712: another Q.

If we want to enable user to write "anything" they want, the port mapping will obstruct them. Should we change the network mode from bridge to host?

saivirtue: streamApp的溝通都是透過broker，使用者會用到port mapping的機會不大。假如有我遺漏的地方可以再討論，目前我是比較傾向將host與container的網路隔開

chia7712: >> 目前我是比較傾向將host與container的網路隔開

It is ok to me. But it would be better to document this

saivirtue: sure.

saivirtue:
retry

PR #375, opened 2019-03-06
OHARA-1668 deps.dockerfile is broken since TestDatabaseClient was mov…

chia7712:
…ed from ohara-it to ohara-client


chia7712:
Test this pr on local. works well :)

PR #388, opened 2019-03-07
#387 Implement a simple cache to speed up ssh collie

### LINKED ISSUE #387: Implement a simple cache to speed up ssh collie
SSH collie relies on ssh connection to control remote node. It is simple but slow. Normally, most ops on collie is readonly, so it worth having a cache layer to speed up the read ops.

chia7712:
required by #387


chia7712:
TestK8sClusterCollie is traced by #393

TestSshClusterCollie pass. see https://builds.is-land.com.tw:59478/job/PreTest-OHARA/270/console

PR #394, opened 2019-03-08
#181 Instead of putting all module into single assembly, we should ge…

chia7712:
required by #181


chia7712:
failed tests are traced by #393

Issue #400, opened 2019-03-09
Apply com.island.ohara.common.annotations.Optional to all builder pattern


chia7712:
we had introduced Optional to help user to build object through build pattern. However, the concept is not applied to all builder in ohara.

The builders should be fixed are shown below.

1. Consumer.Builder (#401)
1. Producer.Builder and Producer.Sender (#409)
1. TopicCreator (#410)

chia7712:
All PRs are completed. close this issue.

PR #401, opened 2019-03-09
#400 Apply com.island.ohara.common.annotations.Optional to Consumer.B…

chia7712:
required by #400


chia7712:
TestDockerClient.testPortMapping is traced by #405 

TestK8sClusterCollie is traced by #393

Issue #402, opened 2019-03-09
Swallow InterruptedException caused by closing cluster cache


chia7712:
the exception is too noisy... 

```
    private[this] val executor = {
      val exec = Executors.newSingleThreadExecutor()
      exec.execute(() =>
        try {
          while (!isClosed) try {
            // we ignore the returned value
            requests.poll(expiredTime.toMillis, TimeUnit.MILLISECONDS)
            // make all thread to wait updating of cache
            lastUpdate = -1
            // TODO: 30 seconds should be enough to fetch result ?
            obj = Await.result(updater(), 30 seconds)
            lastUpdate = CommonUtil.current()
          } catch {
            case e: InterruptedException =>
              // we are closing this thread
              throw e // -----------------------------------------------> this one
            case e: Throwable =>
              LOG.error("failed to update cache", e)
          } finally {
            lock.synchronized {
              lock.notifyAll()
            }
            requests.clear()
          }
        } finally LOG.info(s"cache thread is gone"))
      exec
    }
```

jellynina:
@chia7712 Could I take over this issue?

chia7712:
> Could I take over this issue?

sure.

PR #406, opened 2019-03-10
#402 Swallow InterruptedException caused by closing cluster cache

jellynina:
required by #402


chia7712:
retry it

chia7712:
The failure is traced by #393

Issue #303, opened 2019-02-21
Refactor docker client


saivirtue:
required by #427

This issue consists of three PRs.
1. Extract ContainerCreator from DockerClient and introduce FakeDockerClient (#440)
1. Refactor ContainerCreator so as to reuse most of code in FakeDockerClient and DockerClientImpl   (#470)
1. Complete all tests for ContainerCreator (#472)

saivirtue:
required by #427

saivirtue:
required by #427

saivirtue:
required by #427

saivirtue:
Since some UT tests use DockerClient requiring installed docker, we refactor the DockerClient to support such UT cases.

chia7712:
@saivirtue Good patch. Let me address more refactor.

chia7712:
All PRs are completed. close this issue now.

PR #440, opened 2019-03-12
#303 refactor docker client

saivirtue:
### Checklist
- [x] Set a Milestone (we usually file PR for master branch)
- [x] Link to an issue(You can do so by using the required by keyword: required by #303)


ohara-configurator/src/main/scala/com/island/ohara/configurator/fake/FakeDockerClient.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  private[this] def listContainers(nameFilter: String => Boolean): Seq[ContainerInfo] =
+    cachedContainers.filter { case (name, _) => nameFilter(name) }.values.toSeq
+
+  override def containers(nameFilter: String => Boolean): Seq[ContainerInfo] = listContainers(nameFilter)
+
+  //there is no meaning of "active" in fake mode
+  override def activeContainers(nameFilter: String => Boolean): Seq[ContainerInfo] = listContainers(nameFilter)
+
+  override def containerCreator(): ContainerCreator = new ContainerCreator {
```
chia7712: There are two implementations of ContainerCreator (another is in DockerClientImpl) and they have a bunch of similar code. Could we make ContainerCreator be a super class with the configured value rather than a pure interface?  BTW, it would be better to check the input arguments by the setter.

saivirtue: > BTW, it would be better to check the input arguments by the setter.

good point. Try to refactor `ContainerCreator`

saivirtue:
retry it

saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
```diff
@@ -0,0 +1,221 @@
...
+    */
+  protected[agent] def dockerCommand(): String
+}
+
+object ContainerCreator {
+
+  def builder(agent: Agent): Builder = new Builder(agent)
+
+  class Builder(agent: Agent) {
+    var hostname: String = _
```
chia7712: private[this]

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+      * set container's name. default is a random string
+      *
+      * @param name container name
+      * @return this builder
+      */
+    @Optional("default is CommonUtil.randomString()")
+    def name(name: String): Builder = {
+      this.name = name
```
chia7712: check the input argument. Consider using [CommonUtil.requireNonEmpty](https://github.com/oharastream/ohara/blob/master/ohara-common/src/main/java/com/island/ohara/common/util/CommonUtil.java#L379)

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
```diff
@@ -0,0 +1,221 @@
...
+    * @return the command used to start docker container
+    */
+  protected[agent] def dockerCommand(): String
+}
+
+object ContainerCreator {
+
+  def builder(agent: Agent): Builder = new Builder(agent)
+
+  class Builder(agent: Agent) {
```
chia7712: I don't catch the point of passing agent to builder. The agent is bound with a remote node(hostname, port, user, and password). However, we allow caller to define the information again?

saivirtue: builder need to "build" the container by the agent. what's the caller here meaning?

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
saivirtue: OK, I will modify the checker method inside `def build()`

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
saivirtue: the `ContainerCreator.Builder` class here is used for the both `agent` and `configurator` modules...I am not sure if there was some ways to hide this builder class?

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
chia7712: 如果可以的話，盡量把agent隱藏起來。對於使用dockerclient的人(caller)應該是看不到agent

階層關係大概是  caller -> docker client -> agent -> ssh client

也就是compose的概念

PR #470, opened 2019-03-14
#303 Refactor ContainerCreator

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #303)


chia7712:
FakeDockerClient has a bunch of code similar to ContainerCreator. This PR move all setter to ContainerCreator so other sub classes can reuse them.

chia7712:
@saivirtue Could you file another PR to complete tests of TestContainerCreator if you have free cycle?

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
```diff
@@ -18,56 +18,103 @@ package com.island.ohara.agent.docker
...
+  protected def doExecute(hostname: String,
+                          imageName: String,
+                          name: String,
+                          command: String,
+                          removeContainerOnExit: Boolean,
+                          ports: Map[Int, Int],
+                          envs: Map[String, String],
+                          route: Map[String, String],
+                          volumeMapping: Map[String, String],
+                          networkDriver: NetworkDriver): Unit
```
saivirtue: I see...the point to reuse builder-pattern is put the "builder" part in interface and pass the builder params to the "override" sub-class, right?

saivirtue:
> @saivirtue Could you file another PR to complete tests of TestContainerCreator if you have free cycle?

sure thing, I will add later.

ohara-agent/src/main/scala/com/island/ohara/agent/docker/ContainerCreator.scala
chia7712: > the point to reuse builder-pattern is put the "builder" part in interface and pass the builder params to the "override" sub-class, right?

yep. However, this is just a variant of builder pattern. Not all use cases are suitable to this pattern :)

chia7712:
Unrelated error. Will commit this PR

saivirtue:
> Unrelated error. Will commit this PR

should I add commit to this branch?

chia7712:
> should I add commit to this branch?

File another PR connected to #303 . Refactor is a big news worth having three pull requests :)

Issue #329, opened 2019-02-25
The "state" should be consistent between all components


saivirtue:
目前Configurator內的狀態(state)會有兩種
* ConnectorState
* ContainerState
在Configurator上需要統一兩種狀態對外的response state (ObjectState)

預計工作：
- [x] Add ObjectState as the top level state class (#329)
- [ ] Change all sealed class to extend Enum class (#525)

PR #485, opened 2019-03-19
#329 The state should be consistent between all components

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #329)


saivirtue:
retry

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ConnectorApi.scala
```diff
@@ -29,6 +30,30 @@ object ConnectorApi {
   val PAUSE_COMMAND: String = "pause"
   val RESUME_COMMAND: String = "resume"
 
+  trait ConnectorState { val name: String }
+  object ConnectorState extends Enum[ConnectorState] {
```
chia7712: in scala I prefer to use sealed case class which is more simple and is able to be checked during compile.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,47 @@
...
   val UNKNOWN: String = "?"
   val PIPELINES_PREFIX_PATH: String = "pipelines"
+
+  /**
+    * Represent an "interface" for the pipeline state.
+    * This is the top level class if there was any other kinds of state that are used in the pipeline.
+    *
+    * @param name the actual name of the sate
+    */
+  abstract class ConfiguratorState(name: String)
```
chia7712: It seems to me ConfiguratorState is about "configurator" rather than connector (or streamapp).

How about ObjectState?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ConnectorApi.scala
saivirtue: OK, I will change the `trait` to `abstract sealed class`

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
saivirtue: Hmm...How about PipelineState ? Since pipeline is "top-level" view for all other components

saivirtue:
> Please add test of serializing the new "State" to json string.

Sure.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
chia7712: PipelineState seems be related to pipeline's state.

We should have a super class of all "object" in pipeline. There is already a "Data" in ohara-configurator. Perhaps DataState?

jackyoh:
retry

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ContainerApi.scala
```diff
@@ -58,21 +55,17 @@ object ContainerApi {
...
-      CREATED,
-      RESTARTING,
-      RUNNING,
-      REMOVING,
-      PAUSED,
-      EXITED,
-      DEAD
-    )
-
+  object K8sContainerState {
```
chia7712: Consider adding a helper method to convert string to K8sContainerState.

For example: valueOf(String): K8sContainerState

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -31,11 +65,10 @@ object PipelineApi {
   final case class ObjectAbstract(id: String,
                                   name: String,
                                   kind: String,
-                                  state: Option[ConnectorState],
+                                  state: Option[PipelineState],
```
chia7712: As discussed early, adding PipelineState to a object which is not "Pipeline" may confuse us.

It is great to have a parent class to show the state of all object in pipeline. However, the PipelineState is not a good naming I'd say...

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ContainerApi.scala
saivirtue: Since K8s states need to reflect after K8s feature completed (#429). But that will be ok to extend the k8s part with Enum as others. Thanks!

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
saivirtue: Well, maybe change to ObjectState is good for now (will leave some TODO in code)

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ContainerApi.scala
chia7712: I guess #429 won't be addressed recently. Please fell free to refactor k8s code.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
chia7712: If ObjectState is ok to you, just address it in this PR.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,49 @@
...
+    * @param name the actual name of the sate
+    */
+  abstract class ObjectState(name: String)
+  object ObjectState extends Enum[ObjectState] {
+    def convert(name: String): ObjectState = {
+      all.find(_.name == name).getOrElse(throw new IllegalArgumentException(s"Unexpected name : $name"))
+    }
+    //TODO : Is there a way to remove this redundant code?...by Sam
+    // Connector State Part
+    case object UNASSIGNED extends ObjectState("UNASSIGNED")
```
chia7712: If ObjectState is parent of all state now, the sub class should reference the "name" from ObjectState.

For example

case object CREATED extends ContainerState(ObjectState.CREATED.name)

ohara-client/src/main/scala/com/island/ohara/client/configurator/ConfiguratorApiInfo.scala
```diff
@@ -32,4 +33,26 @@ object ConfiguratorApiInfo {
...
+        .map(symbol => mirror.reflectModule(symbol.asModule).instance)
+        .collect { case v: Enu => v }
+        .toSeq
+        .sortBy(_.name)
+    }
+
+    def forName(name: String): Enu =
+      all.find(_.name == name).getOrElse(throw new IllegalArgumentException(s"Unexpected value : $name"))
+
+    implicit class RichardEnum(enum: Enu) {
```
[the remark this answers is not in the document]
saivirtue: this implicit is used to "trick" name resolving for enum -> string in subclass, hence cannot be private

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
saivirtue: I am not sure it is a good idea to extend a "sealed" class that is represent enumeration...

chia7712: > I am not sure it is a good idea to extend a "sealed" class that is represent enumeration...

We don't need to extend a sealed class from another class file :) Referencing the "constant" is good enough.

ohara-client/src/main/scala/com/island/ohara/client/configurator/ConfiguratorApiInfo.scala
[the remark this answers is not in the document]
saivirtue: I think there is no need to use this trick...have remove now

saivirtue:
retry agent

saivirtue:
retry

saivirtue:
retry configurator

saivirtue:
retry

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,52 @@
...
+    * @param name the actual name of the sate
+    */
+  abstract sealed class ObjectState { val name: String }
+  object ObjectState extends Enum[ObjectState] {
+    def convert(name: String): ObjectState = {
+      all.find(_.name == name).getOrElse(throw new IllegalArgumentException(s"Unexpected name : $name"))
+    }
+    //TODO : Is there a way to remove this redundant code?...by Sam
+    // Connector State Part
+    case object UNASSIGNED extends ObjectState { val name: String = ConnectorState.UNASSIGNED.name }
```
chia7712: 不太確定我腦袋裡的階層有沒有和你的同步： 
ObjectState should be the top of state. Hence, ObjectState includes all possible state from other "states". That is to say, it is wrong to make ObjectState reference to sub state.

saivirtue: 首先，這裡是一個很醜的寫法...如果可以我是希望使用類似下面的寫法：
ConnectorState.all.foreach(state => {initial all values inside ObjectState} )
Anyway, 這裡的意思是不是應該要讓其他的元件去"externd" ObjectState, 而不是在ObjectState內引用其他人?

chia7712: > ConnectorState.all.foreach(state => {initial all values inside ObjectState} )

首先，如果這邊的概念是"列舉"的話，就不適合動態產生

> Anyway, 這裡的意思是不是應該要讓其他的元件去"externd" ObjectState, 而不是在ObjectState內引用其他人?

就階層的角度來說，要讓sub class去引用parent。由於這邊的設計是ObjectState包含所有"可以"發生的狀態，而sub state則都是ObjectState的子集合，因此sub state要使用的"狀態名稱"就應該參考至ObjectState

saivirtue: > 由於這邊的設計是ObjectState包含所有"可以"發生的狀態，而sub state則都是ObjectState的子集合

          ObjectState
            /           \  
          /                \
ConnectorState  ContainerState

got the point , fix right away.

chia7712: It seems to me creating a contains-all-state object MAY be not acceptable since it is not easy to keep compatibility in the future. For example, we want to add another "object" to pipeline and the object is not based on connector or container. Or we want to extend the stats of connector/container and then we have to add the state to parent (objectstate). Both cases cause that all sub state will be impacted by a bit change.

Why not just define the state for each "object" instead of extracting a "parent" for them? In converting to json we can use "string" to replace the state easily.

saivirtue:
@chia7712 
have a free time to take a look for the ObjectState changes?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,50 @@
...
+    * This is the top level class if there was any other kinds of state that are used in the pipeline.
+    * TODO : "Pipeline API" should only see the base states (ex: RUNNING, STOP, FAILED), we could abstract these states
+    * and move to pipeline level "pipelineState"...by Sam
+    *
+    * @param name the actual name of the sate
+    */
+  abstract sealed class ObjectState { val name: String }
+  object ObjectState extends Enum[ObjectState] {
+    def convert(name: String): ObjectState = {
+      all.find(_.name == name).getOrElse(throw new IllegalArgumentException(s"Unexpected name : $name"))
```
chia7712: NoSuchElementException

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,50 @@
...
+    case object FAILED extends ObjectState { val name: String = "FAILED" }
+    case object DESTROYED extends ObjectState { val name: String = "DESTROYED" }
+    case object CREATED extends ObjectState { val name: String = "CREATED" }
+    case object RESTARTING extends ObjectState { val name: String = "RESTARTING" }
+    case object REMOVING extends ObjectState { val name: String = "REMOVING" }
+    case object EXITED extends ObjectState { val name: String = "EXITED" }
+    case object DEAD extends ObjectState { val name: String = "DEAD" }
+  }
+
+  implicit val PIPELINE_STATE_JSON_FORMAT: RootJsonFormat[ObjectState] =
```
chia7712: PIPELINE_STATE_JSON_FORMAT -> OBJECT_XXX

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,50 @@
...
+    * Represent an "interface" for all kinds of state.
+    * This is the top level class if there was any other kinds of state that are used in the pipeline.
+    * TODO : "Pipeline API" should only see the base states (ex: RUNNING, STOP, FAILED), we could abstract these states
+    * and move to pipeline level "pipelineState"...by Sam
+    *
+    * @param name the actual name of the sate
+    */
+  abstract sealed class ObjectState { val name: String }
+  object ObjectState extends Enum[ObjectState] {
+    def convert(name: String): ObjectState = {
```
chia7712: the super class already has forName() so why you create another similar function?

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
```diff
@@ -129,12 +129,14 @@ private[configurator] object PipelineRoute {
...
-                                     name = data.name,
-                                     kind = data.kind,
-                                     state = state,
-                                     error = ConnectorRoute.errorMessage(state),
-                                     lastModified = data.lastModified)
+                      ObjectAbstract(
+                        id = data.id,
+                        name = data.name,
+                        kind = data.kind,
+                        state = state.map(v => ObjectState.convert(v.name)),
```
chia7712: forName

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
```diff
@@ -172,7 +174,13 @@ private[configurator] object PipelineRoute {
               }
 
           case data: StreamApp =>
-            Future.successful(ObjectAbstract(data.id, data.name, data.kind, data.state, None, data.lastModified))
+            Future.successful(
+              ObjectAbstract(data.id,
+                             data.name,
+                             data.kind,
+                             data.state.map(v => ObjectState.convert(v.name)),
```
chia7712: forName

ohara-client/src/main/scala/com/island/ohara/client/kafka/Enum.scala
```diff
@@ -0,0 +1,50 @@
...
+      .toSeq
+      .sortBy(_.toString)
+  }
+
+  // convert string to enum class.
+  def forName(name: String): Enu =
+    all
+      .find(_.toString == name)
+      .getOrElse(
+        throw new IllegalArgumentException(s"Unexpected value : $name")
```
chia7712: NoSuchElementException

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
```diff
@@ -15,13 +15,50 @@
...
+    case object RESTARTING extends ObjectState { val name: String = "RESTARTING" }
+    case object REMOVING extends ObjectState { val name: String = "REMOVING" }
+    case object EXITED extends ObjectState { val name: String = "EXITED" }
+    case object DEAD extends ObjectState { val name: String = "DEAD" }
+  }
+
+  implicit val PIPELINE_STATE_JSON_FORMAT: RootJsonFormat[ObjectState] =
+    new RootJsonFormat[ObjectState] {
+      override def write(obj: ObjectState): JsValue = JsString(obj.name)
+      override def read(json: JsValue): ObjectState = ObjectState.all
```
chia7712: ObjectState.forName

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/PipelineApi.scala
saivirtue: miss the forName function...

saivirtue:
retry

chia7712:
@saivirtue Thanks for this great PR.

Issue #430, opened 2019-03-11
Implement StreamCollie


saivirtue:
required by #427

saivirtue:
This issue has following tasks to be done:
- [ ] Introduce container cluster concept : Crane (WIP #430)
- [] Implement StreamCrane

PR #541, opened 2019-03-25
#430 Implement container cluster : Crane

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #430)


saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
```diff
@@ -0,0 +1,121 @@
...
+import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future}
+import scala.reflect.runtime.universe._
+
+/**
+  * Warehouse is a factory to store all cargoes.
+  * This class help to manage the container group.
+  */
+trait Warehouse {
+  type T <: DockerClient
```
chia7712: Why we have to define a DockerClient for Warehouse explicitly?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  override def build(): Cargo[FakeDockerClient] =
+    new CargoImpl[FakeDockerClient](CommonUtils.requireNonEmpty(_name), Objects.requireNonNull(_node)) {
+
+      override def cargoName: String = _name
+
+      override def node: Node = _node
+
+      override def client(): FakeDockerClient = new FakeDockerClient
```
chia7712: Can FakeDockerClient be closed multi times?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
```diff
@@ -0,0 +1,152 @@
...
+    new CargoImpl[DockerClient](CommonUtils.requireNonEmpty(_name), Objects.requireNonNull(_node)) {
+
+      override def cargoName: String = _name
+
+      override def node: Node = _node
+
+      override def client(): DockerClient = doGet(node: Node)
+
+      private[this] def doGet(node: Node): DockerClient = {
+        Objects.requireNonNull(node)
```
chia7712: Is the input node different to node hosted by CargoDockerBuilder? If not, why require the node again from input argument?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Crane.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+    * Add a warehouse with required size
+    *
+    * @param size the size of warehouse (will create equal size cargo)
+    * @param nodeCollie the NodeCollie that will be used to create warehouse
+    * @param context execution context
+    * @return the added warehouse
+    */
+  def add(size: Int)(implicit nodeCollie: NodeCollie, context: ExecutionContext): Future[Warehouse]
```
chia7712: I guess you plan to get node connection info from NodeCollie. how to get the connection should be defined by implementation rather than interface.

ohara-agent/src/main/scala/com/island/ohara/agent/docker/fake/FakeDockerClient.scala
```diff
@@ -14,21 +14,20 @@
  * limitations under the License.
  */
 
-package com.island.ohara.configurator.fake
+package com.island.ohara.agent.docker.fake
```
vitojeng: Does we need a new *fake* package under *com.island.ohara.agent.docker*?
Or just simply put FackDockerClient into *com.island.ohara.agent.docker*.

chia7712:
Another question. Do you try to hide the "container" from crane and other new class? If so, it would be better to remove "DockerClient" from builder since the existence of DockerClient is related to the implementation.

ohara-agent/src/main/scala/com/island/ohara/agent/docker/fake/FakeDockerClient.scala
saivirtue: 這裡想要把`fake`類的物件都獨立出來 (如`com.island.oharaconfigurator.fake`
以目前的情形來說，的確只會有一個`FakeDockerClient`在裡面

vitojeng: 若是如此, fake package 是否應該為 **com.island.ohara.agent.fake** ?

saivirtue: OK~

saivirtue:
> Just a suggestion. Writing a example to display the relationship between all new class. DON'T create a new class without any usage...

I will add tests to clarify this

saivirtue:
> Another question. Do you try to hide the "container" from crane and other new class? If so, it would be better to remove "DockerClient" from builder since the existence of DockerClient is related to the implementation.

DockerClient is used to provide type for Cargo and other classes. I try to connect cargo <-> DockerClient in the Cargo.Builder, could you give more clues about the "remove DockerClient"?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
```diff
@@ -0,0 +1,152 @@
...
+    * @return node
+    */
+  def node: Node
+
+  /**
+    * Get the client interface to interact container
+    *
+    * @return the client
+    */
+  def client(): T
```
chia7712: 我現在很納悶的地方是為何要讓Cargo有能力去"建立"一個需要被release的物件？Cargo從字面是某種pojo，而DockerClient是一個"很底層"的操作，你封裝了一大堆界面但最後傳回一個很底層的界面...還是你預期使用者拿到Cargo後會"自行"操作DockerClient?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
chia7712: 你兩個實作都會回傳DockerClient。而我看Cargo並沒有要管理dockerclient的意圖，換言之你應該預期caller要自己管理dockerCLient。

可是一邊的client()永遠回傳同一個instance，另一邊則每次都回傳新的instance。對於caller而言拿到的dockerclient到底可不可以關閉？

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Crane.scala
chia7712: 我的意思是你在界面上宣告了NodeCollie，等同於強迫所有impl都必須從NodeCollie取得node的資訊，也就是你在"界面上"已經決定了實作的細節，這樣的抽象是不適當的

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
saivirtue: 這裡的確我期望Cargo不去管理DockerClient，只負責"提供"操作的介面
但這裡的"使用者"會去操作Cargo應該只有Warehouse，我會把Cargo改成package access

saivirtue:
> 你兩個實作都會回傳DockerClient。而我看Cargo並沒有要管理dockerclient的意圖，換言之你應該預期caller要自己管理dockerCLient。
可是一邊的client()永遠回傳同一個instance，另一邊則每次都回傳新的instance。對於caller而言拿到的dockerclient到底可不可以關閉？

docker client是不可以關閉的，但這邊提醒了一件事：我要把DockerClient reuse. 這邊我在來改一下

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
saivirtue: 這裡DockerClient提供Warehouse一個type : 讓上層呼叫者可以決定Warehouse是管理哪種container group (docker , k8s, or fake)

chia7712: > 決定Warehouse是管理哪種container group (docker , k8s, or fake)

用flag表示會遠比用程式碼來得好...oo的一個大原則，界面應該暴露越少的細節越好...

chia7712:
再提醒一下，DockerClient是一個非常非常底層APIs。在定義"界面"的時候不太適合直接把它擺出來

chia7712:
> docker client是不可以關閉的，但這邊提醒了一件事：我要把DockerClient reuse. 這邊我在來改一下

你如果要實作"reuse"，那麼就要有個傢伙來負責管理（關閉）dockerclient

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClientCache.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.island.ohara.agent.ssh.DockerClientCache
+import com.island.ohara.client.configurator.v0.NodeApi.Node
+import com.island.ohara.common.util.{Releasable, ReleaseOnce}
+
+import scala.collection.mutable
+
+/**
+  * Similar to DockerClientCache, but fake
+  */
+private[agent] trait FakeDockerClientCache extends DockerClientCache
```
chia7712: DockerClientCache is a functional interface. Hence, you can implement it by lambda expression.

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClientCache.scala
```diff
@@ -0,0 +1,51 @@
...
+
+private[agent] object FakeDockerClientCache {
+  def apply(): FakeDockerClientCache = new DockerClientCacheImpl()
+
+  private[this] class DockerClientCacheImpl extends ReleaseOnce with FakeDockerClientCache {
+    private[this] val lock = new Object()
+    private[this] val cache: mutable.HashMap[Node, FakeDockerClient] = new mutable.HashMap[Node, FakeDockerClient]()
```
chia7712: I DON'T think fake docker client still need cache...

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClientCache.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.island.ohara.common.util.{Releasable, ReleaseOnce}
+
+import scala.collection.mutable
+
+/**
+  * Similar to DockerClientCache, but fake
+  */
+private[agent] trait FakeDockerClientCache extends DockerClientCache
+
+private[agent] object FakeDockerClientCache {
```
chia7712: Please try to "reduce" the duplicate code. Clone with "a bit" change is always not a good style...

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Crane.scala
```diff
@@ -0,0 +1,65 @@
...
+  /**
+    * Delete a warehouse
+    *
+    * @param name warehouse name
+    * @param executionContext execution context
+    * @return the deleted cargo group, or throw exception if not exists
+    */
+  def delete(name: String)(implicit executionContext: ExecutionContext): Future[Warehouse]
+}
+object Crane {}
```
chia7712: ?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
```diff
@@ -0,0 +1,79 @@
...
+    * Delete cargo by it's name
+    *
+    * @param cargoName the cargo name that is want to be deleted
+    * @param executionContext execution context
+    * @return the deleted cargo, or exception if no such cargo
+    */
+  def delete(cargoName: String)(implicit executionContext: ExecutionContext): Future[Cargo]
+}
+
+object Warehouse {}
```
chia7712: ?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
```diff
@@ -0,0 +1,143 @@
...
+
+sealed abstract class StreamCargo extends Cargo {
+  type T = ContainerInfo
+
+  /**
+    * single entry to build the stream container (use docker by default)
+    *
+    * @return the builder
+    */
+  def builder(): StreamCargo.Builder
```
chia7712: StreamCargo.Builder => it implies the builder is used to build StreamCargo. see google style (https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Message.Builder)

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
```diff
@@ -0,0 +1,85 @@
...
+import com.island.ohara.common.util.CommonUtils
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * Cargo contains all kind of goods and ready to be conveyed.
+  * We store the container information in this class.
+  */
+trait Cargo {
+  type T <: ContainerInfo
```
chia7712: why not using generic directly?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
```diff
@@ -0,0 +1,85 @@
...
+    * @return the container info
+    */
+  def info(): Future[T]
+
+  /**
+    * Remove this cargo (and the dependent container)
+    *
+    * @return true if the cargo was removed, false otherwise
+    */
+  def remove(): Future[Boolean]
```
chia7712: we allow a object to suicide? how to check the pulse if I get a Cargo?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    override def builder(): Builder = new Builder(name, client)
+
+    override def info(): Future[ContainerInfo] = {
+      Future.successful(client.container(name))
+    }
+
+    override def remove(): Future[Boolean] = {
+      try {
+        client.forceRemove(name)
```
chia7712: forceRemove do send kill -9 and you should have a way to stop container gracefully...

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
saivirtue: it's another style for generic. we can hide this information from class definition

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
saivirtue: that make sense. I will change the name to be more "not like suicide"

saivirtue: this function is only responsible for "container remove"

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
saivirtue: OK, let me add flag to switch graceful and force

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClientCache.scala
saivirtue: refactor now

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
chia7712: I love this good mechanism. We can apply it to code base.

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
```diff
@@ -0,0 +1,79 @@
...
+    list().map(_.find(_.name == cargoName))
+
+  /**
+    * Add a cargo to this warehouse (should also
+    *
+    * @param node the node to be added cargo
+    * @param executionContext execution context
+    * @return value of the added cargo if not exists, or the previous value
+    */
+  def add(node: Node)(implicit executionContext: ExecutionContext): Future[Cargo]
```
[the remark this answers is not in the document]
saivirtue: crane does not care the `Node` informatin ; it is only used to manage warehouse.
but in implementation, when caller is in view of "crane", the node information should be passed from crane to warehouse , hence the node in warehouse should be subset of node in crane

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
chia7712: Another Q: Is it necessary to abstract "ContainerInfo"? Do we have sub class of ContainerInfo in the future?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
```diff
@@ -0,0 +1,143 @@
...
+  /**
+    * single entry to build the stream container (use docker by default)
+    *
+    * @return the builder
+    */
+  def builder(): StreamCargo.Builder
+}
+private[stream] object StreamCargo {
+
+  def apply(client: DockerClient): StreamCargo = new StreamCargo {
```
chia7712: You keep a "possible" shared object in this object. How you control the close flow? Release all cargos first and the close all docker client?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Crane.scala
```diff
@@ -0,0 +1,65 @@
...
+    * @param executionContext execution context
+    * @return the required warehouse, or throw exception if not exists
+    */
+  def get(name: String)(implicit executionContext: ExecutionContext): Future[Warehouse] =
+    list().map(
+      _.find(_.name == name)
+        .getOrElse(throw new IllegalArgumentException(s"no such warehouse by require name : $name")))
+
+  /**
+    * Add a warehouse with required size (the cargo will be created also)
```
chia7712: > the cargo will be created also

the container will be created also?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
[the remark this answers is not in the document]
chia7712: > but in implementation, when caller is in view of "crane", the node information should be passed from crane to warehouse , hence the node in warehouse should be subset of node in crane

this is already implemented?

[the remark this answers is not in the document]
chia7712: As a interface method, requiring user to input "Node" is a little weird. Since, why callers has to "know" the password/port/username when they "just" want to create a cargo on a specific node?

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
saivirtue: Not sure if k8s has more information in the future...I keep the flexible

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Warehouse.scala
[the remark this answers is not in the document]
saivirtue: > this is already implemented?

See `StreamCrane`

> As a interface method, requiring user to input "Node" is a little weird.

Hmm...that make sense. I will change this to `nodeName` instead

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
chia7712: ContainerInfo is case clas...I DON'T think use a "case class" in abstract type is a good case.

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
saivirtue: a cargo own one client.
warehouse keep the "client cache" ; hence the flow will look like :
warehouse delete a cargo -> container delete first -> cargo delete -> docker client will keep for that node for further usage

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Crane.scala
saivirtue: yes~ let me add more desc

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/Cargo.scala
saivirtue: OK, I will change this to `type T = ContainerInfo`

chia7712: Could we just get rid of abstract type? Code should be a document also. Abstracting a type which WON'T be "abstracted" is weird.

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    override def createContainer(): ContainerCreator = new ContainerCreator(name, client)
+
+    override def info(): Future[ContainerInfo] = {
+      Future.successful(client.container(name))
+    }
+
+    override def removeContainer(isForce: Boolean): Future[Boolean] = {
+      try {
+        if (isForce) client.forceRemove(name)
```
chia7712: chain the future please

ohara-agent/src/main/scala/com/island/ohara/agent/wharf/stream/StreamCargo.scala
chia7712: Do you have a rule to cleanup all cargo in graceful shutdown??

chia7712:
@saivirtue Consider the deadline. You can merge this PR after passing QA.

saivirtue:
> @saivirtue Consider the deadline. You can merge this PR after passing QA.

OK, let me resolve the chain future part and add some comment

Issue #637, opened 2019-04-03
Refactor CommonUtils.requireNonEmpty


saivirtue:
Currently, we have following function to check object is `not null or empty` :

1. requireNonEmpty(**String** s)
2. requireNonEmpty(**String** s, Supplier<String> msg)
3. requireNonEmpty<**Collection** T>(Ts)
4. requireNonEmpty<**Collection**>(Ts, Supplier<String> msg)
5. requireNonEmpty<**Map**>(T s)
6. requireNonEmpty<**Map**>(T s, Supplier<String> msg)

But we could not write the following code in scala :
`requireNonEmpty(seq: Seq[String])`

this issue will address the problem and refactor utils

saivirtue:
@chia7712 
因為目前很多scala都使用自己的方式去檢查，我會把有使用`requireNonEmpty`的地方都使用CommonUtils

chia7712:
> 因為目前很多scala都使用自己的方式去檢查，我會把有使用requireNonEmpty的地方都使用CommonUtils

great!

PR #663, opened 2019-04-08
#637 Refactor CommonUtils.requireNonEmpty

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #637 )


ohara-common/src/main/java/com/island/ohara/common/util/CommonUtils.java
```diff
@@ -433,11 +431,23 @@ public static String requireNonEmpty(String s, Supplier<String> msg) {
    * @throws IllegalArgumentException if {@code s} is empty
    * @return input map
    */
-  public static <T extends Map<?, ?>> T requireNonEmpty(T s, Supplier<String> msg) {
+  public static <T extends Map<?, ?>> T requireNonEmptyWithMsg(T s, Supplier<String> msg) {
```
chia7712: Why having a new method name??

saivirtue: I found that `scala` will complaint the method arguments is not as expected if we use `requireNonEmpty` for both (T s) and (T s, Supplier<String> msg)...
I think this is a option to solve the problem of java overloading function in scala not really work

chia7712: Could you share a example with me? Thanks!

saivirtue: Uh...I found the problem. Since ohara-common is a base module for almost other modules, I need to `gradle clean` the common module cache. Will change the method name later, thanks !

Issue #601, opened 2019-03-31
Enable metrics on ohara


chia7712:
We already have a great wrap of connector for ohara. As a comprehensive tool, however, we still lack monitor infra - metrics and logs. This issue aims to implements a basic function of metrics that enable row connector to report "the various counter of data flow".

Summarily, there are five sub-tasks are included by this issue.
1. Enable JMX service on configurator (#634)
1. Port of RMI connector needs to be configured also (#635)
1. Add ohara-metrics module, basic bean object and accessor (#657)
1. Enable JMX service on worker cluster (#662)
1. add "unit" to Counter to show the meaning of value in Counter (#667)
1. add background metrics to row connector (#675)
1. the new APIs for /v0/pipelines ([docs](https://paper.dropbox.com/doc/Configurator-APIs--Aa1n6ube4696WXV6y46EHqIiAg-nYWQMqQLGsG0lffMaY7Xk#:uid=953035614408446657328839&h2=[object]-Object-Abstract)) (#676)
1. Add a helper method to create Counter quickly for custom connector (#679)
1. Add metrics to /v0/connector (#684)

2019/04/08
~~a simple APIs of retrieving metrics from row connectors~~
~~a metrics APIs for row counter~~
2019/04/09
~~We have contributed Counter APIs.~~

2019/04/09
Change all metrics methods from blocking to async (movied to issue #678)

chia7712:
Personally, this issue should be incorporated into 0.4. But there is no label for 0.4.

vitojeng:
@chia7712 Can we have a meeting about this issue for more detail idea?

chia7712:
> Can we have a meeting about this issue for more detail idea?

sure

chia7712:
all sub-issues are completed :)

PR #684, opened 2019-04-09
#601 Add metrics to /v0/connector

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #601)


ohara-client/src/test/scala/com/island/ohara/client/configurator/v0/TestConnectorApi.scala
```diff
@@ -212,6 +216,9 @@ class TestConnectorApi extends SmallTest with Matchers {
                                                                                       | \"lastModified\": 123,
                                                                                       | \"settings\": {
                                                                                       | \"className\": \"$className\"
+                                                                                      | },
+                                                                                      | \"metrics\": {
+                                                                                      |   "counters":[]
```
saivirtue: could we show the mbean name here? (ex : MBean: java.lang:type=ClassLoading)

chia7712: This test is about parsing the string with "empty metrics" so I put a empty array.

And I'm not sure whether the metrics property key is useful to UI. We can enhance that later :)

Issue #701, opened 2019-04-09
Complete the docs of restful APIs


chia7712:
There are too many APIs in ohara so this issue will be made up of several pull requests.

1. add docs folder and initialize restful apis docs (#704)
1. add objects docs (#713)
1. add connector docs (#716)
1. add pipeline docs (#717)
1. add cluster docs (#725)
1. add validate docs (#737)
1. add containers docs (#739)
1. add streamapp docs (#743)
1. add jars docs (#744)
1. add logs docs (#747)

chia7712:
@vitojeng Shabondi docs seems to be not complete so I don't file a PR for it

vitojeng:
> @vitojeng Shabondi docs seems to be not complete so I don't file a PR for it

Sure. 👌

chia7712:
Except for private APIs, all public APIs have be documented.  Close this issue now.

PR #739, opened 2019-04-12
#701 add containers docs

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x]  DON'T add any labels (PR labels will be added automatically by Jenkins)
- [x] Link to an issue(You can do so by using the required by keyword: required by #701)
see [link](https://github.com/chia7712/ohara/blob/fix_701_6/docs/rest_interface.md)


docs/rest_interface.md
```diff
@@ -2614,6 +2625,85 @@ for you. The element **setting** is what you request to validate.
 1. value (**string**) — the value you request to validate
 1. errors (**array(string)**) — error message when the input value is illegal to connector 
 ----------
+## Container
+
+All process managed by ohara is based on docker container. This APIs provide the container details of a running cluster.
```
saivirtue: a little question, is this implied we will have another doc for k8s container?

chia7712: Not really. ContainerInfo is a part of interface of collie, so both of them are required to return ContainerInof

chia7712: It would be better to keep restful APIs consistent even though we have different implementation in the backend.

saivirtue: got it :)

Issue #527, opened 2019-03-23
JDBC/HDFS/FTP/PERF connectors should get to write their setting definitions


chia7712:
Our metadata of connectors is almost completed. The official connectors should start to write setting definitions which can make user see which settings they should define.

1. Ftp source is traced by #544
1. Ftp sink is traced by #547
1. Hdfs sink is traced by #563 
1. JDBC source is traced by #563
1. Perf source is traced by #791

FYI @jackyoh @konekoya @eechih

chia7712:
@eechih If you have free cycle, please take over PR about adding setting definitions to Perf source

jackyoh:
HDFS Sink Connector 和 JDBC Source Connector 我這邊會做修改

chia7712:
@eechih Do you have free cycle to take over the latest (perf) connector?

eechih:
@chia7712 Yes, I can take over the job.

chia7712:
all sub tasks are completed. close it now.

PR #791, opened 2019-04-17
#527 Add setting definitions to perf source

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #527 )


ohara-connector/src/main/scala/com/island/ohara/connector/perf/PerfSource.scala
```diff
@@ -40,5 +41,23 @@ class PerfSource extends RowSourceConnector {
 
   override protected def _stop(): Unit = {}
 
+  override protected def _definitions(): java.util.List[SettingDefinition] = Seq(
+    SettingDefinition
+      .builder()
+      .displayName("Batch")
+      .documentation("The batch of perf")
+      .valueType(SettingDefinition.Type.INT)
+      .key(PERF_BATCH)
```
chia7712: I feel this should be optional also since it does impact the throughput only.

eechih: I had set the "Batch" definition to optional and default value to 10.

eechih: Please take a look, thanks. @chia7712

PR #792, opened 2019-04-17
#789 streamApp list APIs will fail

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #789)
fail only occure in ohara manager, no need to add tests


chia7712:
Could we have a test for this case?

saivirtue:
> Could we have a test for this case?

For UT tests, I think `TestStreamRoute.scala` has covered most cases.
If we want to see the fail cases in manager, maybe we need to integrate the front-end e2e tests?

chia7712:
I don’t dig in this issue. But is it hard to produce a request similar to https://github.com/akka/akka-http/issues/1216 on our UT?

saivirtue:
> I don’t dig in this issue. But is it hard to produce a request similar to [akka/akka-http#1216](https://github.com/akka/akka-http/issues/1216) on our UT?

這個問題目前要

chia7712:
@saivirtue ok

Issue #751, opened 2019-04-15
change request body of stream APIs


saivirtue:
since we should separate each router's logic, stream router needs to change some API interface.

This issue will change the following APIs :

1. `change` GET /v0/stream/property/${id}

**Response**
```diff
{
+ "workerClusterName": "wk01",
  "id": "d312871a-4a05-488d-aae0-c8b27c5312c2",
  "name": "my-app",
  "instances": 3,
+ "jarInfo": {
+   "id": "1b022c59-93f9-452c-a062-f8e4cb6c00fe",
+   "name": "new-name.jar",
+   "size": 1234,
+   "lastModified": 1542102595892
+ },
  "from": [
    "topic1"
  ],
  "to": [
    "topic2"
  ],
+ "state": "RUNNING",
+ "lastModified": 1542102595892
}
```

2. `change` PUT /v0/stream/property/${id}

***Request***
```diff
{
  "name": "my-new-app",
- "fromTopics": [
+ "from": [
    "topic1"
  ],
- "toTopics": [
+ "to": [
    "topic2"
  ],
  "instances": 3
}
```

**Response**
```diff
{
+ "workerClusterName": "wk01",
  "id": "d312871a-4a05-488d-aae0-c8b27c5312c2",
  "name": "my-app",
  "instances": 3,
+ "jarInfo": {
+   "id": "1b022c59-93f9-452c-a062-f8e4cb6c00fe",
+   "name": "new-name.jar",
+   "size": 1234,
+   "lastModified": 1542102595892
+ },
  "from": [
    "topic1"
  ],
  "to": [
    "topic2"
  ],
+ "state": "RUNNING",
+ "lastModified": 1542102595892
}
```

saivirtue:
ping @konekoya

konekoya:
It looks like this issue doesn't require any work from my part :)

PR #809, opened 2019-04-18
#751 change request body of stream APIs

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #751)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
```diff
@@ -189,18 +192,36 @@ private[configurator] object PipelineRoute {
...
+              crane
+                .get(StreamApi.formatClusterName(data.id))
+                .map(_._1.asInstanceOf[StreamClusterInfo])
+                .map { info =>
+                  ObjectAbstract(
+                    id = data.id,
+                    name = data.name,
+                    kind = data.kind,
+                    className = None,
+                    state = info.state,
```
chia7712: Is this the latest state of streamapp?

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
```diff
@@ -189,18 +192,36 @@ private[configurator] object PipelineRoute {
...
-
+              crane
+                .get(StreamApi.formatClusterName(data.id))
+                .map(_._1.asInstanceOf[StreamClusterInfo])
+                .map { info =>
+                  ObjectAbstract(
+                    id = data.id,
+                    name = data.name,
+                    kind = data.kind,
+                    className = None,
```
chia7712: It would be better to provide the user-defined class in the future.

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
saivirtue: Yes~`crane` will get the latest state of current cluster

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/PipelineRoute.scala
saivirtue: OK. create a new enhance issue #810

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -79,6 +79,34 @@ private[configurator] object StreamRoute {
...
+    * @return updated streamApp data
+    */
+  private[this] def updateState(id: String)(implicit
+                                            store: DataStore,
+                                            crane: Crane,
+                                            executionContext: ExecutionContext): Future[StreamAppDescription] = {
+    store.value[StreamAppDescription](id).flatMap { props =>
+      crane
+        .get(formatClusterName(props.id))
+        .map(_._1.asInstanceOf[StreamClusterInfo].state)
```
chia7712: Please document why it is safe to do a rash casting.

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -79,6 +79,34 @@ private[configurator] object StreamRoute {
...
+        .recover {
+          case ex: Throwable =>
+            log.error(s"failed to fetch stats for $props, maybe cluster not exist?", ex)
+            None
+        }
+        .flatMap { state =>
+          store.update[StreamAppDescription](props.id, previous => Future.successful(previous.copy(state = state)))
```
chia7712: How to tell caller the state is "not" correct (it may be a result of failed fetch) ?

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
saivirtue: OK

jackyoh:
retry

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
saivirtue: let me add some check

Issue #833, opened 2019-04-22
The settings in PerfSource declares they have default value but they are not applied to code base


chia7712:
```java
  override protected def _definitions(): java.util.List[SettingDefinition] = Seq(
    SettingDefinition
      .builder()
      .displayName("Batch")
      .documentation("The batch of perf")
      .valueType(SettingDefinition.Type.INT)
      .key(PERF_BATCH)
      .optional("10")
      .build(),
    SettingDefinition
      .builder()
      .displayName("Frequence")
      .documentation("The frequence of perf")
      .valueType(SettingDefinition.Type.STRING)
      .key(PERF_FREQUENCE)
      .optional("1 second")
      .build(),
  ).asJava
```
The settings returned by PerfSource have defined the default value. But we still throw exception when the input configuration lacks related setting.
```
object PerfSourceProps {
  def apply(props: Map[String, String]): PerfSourceProps = PerfSourceProps(
    batch = props(PERF_BATCH).toInt,
    freq = Duration(props(PERF_FREQUENCE))
  )
}
```

chia7712:
@eechih Do you have free cycle to iron out this issue?

eechih:
> @eechih Do you have free cycle to iron out this issue?

Yes, I have.

PR #848, opened 2019-04-22
#833 The settings in PerfSource declares they have default value but…

eechih:
… they are not applied to code base

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #833 )


ohara-connector/src/main/scala/com/island/ohara/connector/perf/package.scala
```diff
@@ -22,6 +22,9 @@ package object perf {
   val PERF_BATCH: String = "perf.batch"
   val PERF_FREQUENCE: String = "perf.frequence"
 
+  val DEFAULT_BATCH: String = "10"
```
chia7712: It would be better to keep the **original** type, and then convert it to other type according to different use case.

ohara-connector/src/main/scala/com/island/ohara/connector/perf/package.scala
```diff
@@ -22,6 +22,9 @@ package object perf {
...
+  val DEFAULT_FREQUENCE: String = "1 second"
```
chia7712: ditto

ohara-connector/src/main/scala/com/island/ohara/connector/perf/package.scala
eechih: ok

ohara-connector/src/main/scala/com/island/ohara/connector/perf/package.scala
eechih: ok

eechih: 我已經將預設值改成 **original** type，請再幫忙 review 一下，3Q~ @chia7712

eechih:
retry

eechih:
retry

chia7712:
retry

jackyoh:
retry

eechih:
retry

eechih:
retry

eechih:
QA failure for other reasons. Tracking by issue #875

Issue #840, opened 2019-04-22
Add tests for StreamApp APIs


saivirtue:
required by #650

PR #859, opened 2019-04-23
#840 Add tests for StreamApp APIs

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #840 )


ohara-streams/src/main/java/com/island/ohara/streams/data/Stele.java
```diff
@@ -18,23 +18,22 @@
...
 
-@SuppressWarnings("serial")
 public final class Stele extends Data implements Serializable {
+  private static final long serialVersionUID = 1L;
   private final String kind;
   private final String key;
   private final String name;
-  private final List<String> from;
-  private final List<String> to;
+  private final String[] from;
```
chia7712: why?

ohara-streams/src/test/java/com/island/ohara/streams/TestStreamApp.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package com.island.ohara.streams;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestStreamApp {
```
chia7712: It need to extend one of our test catalog (such as SmallTest and MediumTest)

ohara-streams/src/test/java/com/island/ohara/streams/TestStreamAppTopology.java
```diff
@@ -0,0 +1,79 @@
...
+import com.island.ohara.streams.data.Poneglyph;
+import com.island.ohara.streams.data.Stele;
+import com.island.ohara.streams.ostream.KeyValue;
+import com.island.ohara.streams.ostream.Serdes;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestStreamAppTopology {
```
chia7712: ditto

ohara-streams/src/main/java/com/island/ohara/streams/data/Stele.java
saivirtue: We need to use the `ToStringBuilder.reflectionToString` to generate json string, and there is a wield result between `List<String>` and `String[]`
ex : val list = ("a","b")
List<String> : [a,b]
String[] : ["a", "b"]  ,== correct json object

Maybe there is other solution?

ohara-streams/src/test/java/com/island/ohara/streams/TestStreamApp.java
saivirtue: OK

ohara-streams/src/main/java/com/island/ohara/streams/data/Stele.java
chia7712: Seems a bug in Commons lang. Please file a issue for it and then document it in your code.

saivirtue: After a deep trace of `ToStringBuilder`, I think the problem is not related to commons-lang.
detail is in the comment of source code.

chia7712: I'm ok to use array type. Just left a issue for it and document for your code. It won't cause end of world.

saivirtue: see #870

Issue #864, opened 2019-04-23
DataStore#exist() should return false if id not exist


vitojeng:
I expected DataStore#exist() should return `fasle` but throw exception instead.
Am I getting it wrong?

```text
java.util.NoSuchElementException: 54321 doesn't exist. total:7ae12,348b4,24901
	at com.island.ohara.configurator.store.Store$$anon$1.$anonfun$value$2(Store.scala:114)
	at scala.Option.getOrElse(Option.scala:138)
	at com.island.ohara.configurator.store.Store$$anon$1.value(Store.scala:113)
	at com.island.ohara.configurator.store.DataStore$DataStoreImpl.exist(DataStore.scala:105)
	at com.island.ohara.configurator.route.ShabondiRoute$.$anonfun$apply$6(ShabondiRoute.scala:75)
	at akka.http.scaladsl.server.Directive$.$anonfun$addByNameNullaryApply$2(Directive.scala:134)
```

vitojeng:
The current implement:
```scala
    override def exist[T <: Data: ClassTag](id: String)(implicit executor: ExecutionContext): Future[Boolean] =
      store.value(id).map(classTag[T].runtimeClass.isInstance)
```

May should be(?) :
```scala
    override def exist[T <: Data: ClassTag](id: String)(implicit executor: ExecutionContext): Future[Boolean] =
      store.exist(id)
```

chia7712:
> DataStore#exist()

It should return false instead of throwing exception.

```java
    override def exist[T <: Data: ClassTag](id: String)(implicit executor: ExecutionContext): Future[Boolean] =
      store.exist(id)
```

It needs to check the type also.

PR #872, opened 2019-04-23
#864 DataStore#exist() should return false if id not exist

vitojeng:
required by #864

### Checklist
- [ ] [Review Contribution Guidelines](https://github.com/oharastream/ohara/blob/master/CONTRIBUTING.md)
- [ ] Add Reviewers (see [Authors](https://github.com/oharastream/ohara#authors))
- [ ] Add Assignees (of course, you are the assignee by default!)
- [ ] DON'T set a Milestone
- [ ] DON'T add any labels (PR labels will be added automatically by Jenkins)
- [ ] Link to an issue(You can do so by using the required by keyword: required by #issue_number)

### DIFF
--- a/ohara-configurator/src/test/scala/com/island/ohara/configurator/store/TestDataStore.scala
+++ b/ohara-configurator/src/test/scala/com/island/ohara/configurator/store/TestDataStore.scala
@@ -1,0 +1,108 @@
+package com.island.ohara.configurator.store
+
+import com.island.ohara.client.configurator.v0.ConnectorApi.ConnectorDescription
+import com.island.ohara.client.configurator.v0.Data
+import com.island.ohara.common.data.Serializer
+import com.island.ohara.common.rule.MediumTest
+import com.island.ohara.common.util.{CommonUtils, Releasable}
+import com.island.ohara.configurator.Configurator
+import org.junit.{After, Test}
+import org.scalatest.Matchers
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+private[store] final case class TestData(id: String,
+                                          name: String,
+                                          lastModified: Long,
+                                          kind: String) extends Data
+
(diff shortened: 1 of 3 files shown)


chia7712:
@vitojeng Could you add a test ?

vitojeng:
> @vitojeng Could you add a test ?

Sure.

ohara-configurator/src/test/scala/com/island/ohara/configurator/store/TestDataStore.scala
```diff
@@ -0,0 +1,118 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+private[store] final case class TestData(id: String, name: String, lastModified: Long, kind: String) extends Data
+
+class TestDataStore extends MediumTest with Matchers {
+
+  private[configurator] val DATA_SERIALIZER: Serializer[Data] = new Serializer[Data] {
```
chia7712: useless object?

ohara-configurator/src/test/scala/com/island/ohara/configurator/store/TestDataStore.scala
```diff
@@ -0,0 +1,118 @@
...
+import com.island.ohara.common.util.{CommonUtils, Releasable}
+import com.island.ohara.configurator.Configurator
+import org.junit.{After, Test}
+import org.scalatest.Matchers
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+private[store] final case class TestData(id: String, name: String, lastModified: Long, kind: String) extends Data
```
chia7712: Don't use "Test" as prefix for the class which doesn't carry test case.

vitojeng: Got it!

Issue #889, opened 2019-04-23
Fix flaky TestCraneWithoutDockerServer.normalCase


chia7712:
The test case depends on update of cache, and the wait time in the case is not enough.

chia7712:
FYI @saivirtue

PR #890, opened 2019-04-23
#889 Fix flaky TestCraneWithoutDockerServer.normalCase

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #889)
the loop tests are [here](https://builds.is-land.com.tw/job/PreTest-OHARA/342/console)


chia7712:
@saivirtue Please take a look at this RP

ohara-agent/src/test/scala/com/island/ohara/agent/docker/TestCraneWithoutDockerServer.scala
```diff
@@ -74,12 +74,13 @@ class TestCraneWithoutDockerServer extends SmallTest with Matchers {
...
-
-    info._1.name shouldBe WAREHOUSE_NAME
-    info._2.size shouldBe SIZE
+    CommonUtils.await(
+      () => {
+        val info = awaitResult(crane.get(WAREHOUSE_NAME))
+        info._1.name == WAREHOUSE_NAME && info._2.size == SIZE
+      },
+      java.time.Duration.ofSeconds(20)
+    )
```
saivirtue: Does this mean thread will wait till `info` get all the result?

chia7712: yep. If our cache is powerful to update the result quickly, the end of wait ensues. By contrast, we give 20 seconds to wait the **slow** cache. It is still time-based test but is more stable.

saivirtue: so, here the result of `a == b` is equals to `a shouldBe b` as previous?

chia7712: Not really. "a == b" just return a boolean value but "a shouldBe b" produce a exception if comparison result is false.

saivirtue: got it

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClient.scala
```diff
@@ -16,26 +16,32 @@
...
 
-import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
-
+import scala.collection.JavaConverters._
 private[agent] class FakeDockerClient extends ReleaseOnce with DockerClient {
   private val LOG = Logger(classOf[FakeDockerClient])
   private[this] val FAKE_KIND_NAME: String = "FAKE"
-  private[this] val cachedContainers = new mutable.HashMap[String, ContainerInfo]()
+  private[this] val cachedContainers = new ConcurrentHashMap[String, ContainerInfo]()
```
saivirtue: overall +1. 
Another small question, if we use `ConcurrentHashMap`, do we still need using `lock.synchronized` ?

chia7712: > if we use ConcurrentHashMap, do we still need using lock.synchronized ?

If the method operates only the ConcurrentHashMap, we don't need to sync on method. However, the lock is still required when we do multiple operations in the method. For example,
```scala
   val a = ConcurrentHashMap.get(x)
   val b = process(a)
   ConcurrentHashMap.put(b)
``` 

In order to keep data consistency, we ought to add sync block for above case.

```scala
  sync {
    val a = ConcurrentHashMap.get(x)
    val b = process(a)
    ConcurrentHashMap.put(b)
  }
```

chia7712: At any rate, the code are in testing scope. It is not necessary to make a perfect fake stuff :)

Issue #944, opened 2019-04-29
[Docker] hostname of containerCreator should use different name from nodename in stream


saivirtue:
**Describe the bug**
When create a stream cluster, we will assign the `hostname` for each container :
```
containerCreator()
.hostname()
...
```
But the hostname in container should be different from the node name to avoid weird connection refuse in streamApp starting :
```
[Island-StreamApp] INFO org.apache.kafka.clients.consumer.ConsumerConfig - ConsumerConfig values:
        auto.commit.interval.ms = 5000
        auto.offset.reset = earliest
        bootstrap.servers = [ohara-qa2-node01:9092]
...
[Island-StreamApp] INFO com.island.ohara.streams.ostream.OStreamImpl - poneglyph:[{"steles":[{"kind":"Sink","key":"KSTREAM-SINK-0000000002","name":"e77cbfd3-ef44-4b55-a62b-93572a762c8a","from":["KSTREAM-MAPVALUES-0000000001"],"to":[]}, {"kind":"Processor","key":"KSTREAM-MAPVALUES-0000000001","name":"","from":["KSTREAM-SOURCE-0000000000"],"to":["KSTREAM-SINK-0000000002"]}, {"kind":"Source","key":"KSTREAM-SOURCE-0000000000","name":"[d9284b2b-473a-4b04-92c1-7d551ff80fbb]","from":[],"to":["KSTREAM-MAPVALUES-0000000001"]}]}]
[Island-StreamApp] WARN org.apache.kafka.clients.NetworkClient - [StreamsKafkaClient clientId=ost-cf160c5e-4c21-43c7-a0bf-057b1d73d387] Connection to node -1 could not be established. Broker may not be available.
[Island-StreamApp] WARN org.apache.kafka.clients.NetworkClient - [StreamsKafkaClient clientId=ost-cf160c5e-4c21-43c7-a0bf-057b1d73d387] Connection to node -1 could not be established. Broker may not be available.
```

**Expected behavior**
Start normally.

PR #945, opened 2019-04-29
#944 hostname of containerCreator should use different name

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #944)


saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/docker/StreamWarehouseImpl.scala
```diff
@@ -74,8 +74,8 @@ private class StreamWarehouseImpl(nodeCollie: NodeCollie,
                             node,
                             _.containerCreator()
                               .imageName(imageName)
-                              // we should set the hostname to identify container location
-                              .hostname(node.name)
+                              // we should set the hostname to container name
+                              .hostname(name)
```
chia7712: this is indeed a serious bug :(

ohara-agent/src/main/scala/com/island/ohara/agent/fake/FakeDockerClient.scala
```diff
@@ -63,7 +63,7 @@ private[agent] class FakeDockerClient extends ReleaseOnce with DockerClient {
     cachedContainers.put(
       name,
       ContainerInfo(
-        nodeName = hostname,
+        nodeName = nodeName,
```
chia7712: nice

Issue #974, opened 2019-04-30
A generic cache implementation for all slow data operations


chia7712:
We have completed enough functions for ohara and it is time to care for latency. A quick solution is to provide cache for all slow (data) operations, such as topic, pipeline and connector APIs. 

- [x] a generic cache implementation (#985)
- [x] apply cache to topic APIs (#992)
- [x] apply cache to connector APIs (#997)
- [x] apply cache to pipeline APIs (#1003)
- [x] apply cache to collie (#1019)

#1023 is the follow-up which removes ssh.Cache actually.

saivirtue:
> a generic cache implementation
Will this interface be placed in `common` module?

chia7712:
> Will this interface be placed in common module?

Yep. It will replace all cache code in ohara

PR #985, opened 2019-05-02
#974 a generic cache implementation

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #974)


gradle/dependencies.gradle
```diff
@@ -42,6 +42,8 @@ versions["commonNet"] = project.hasProperty('commonNet.version') ? project.prope
 versions["sshd"] = project.hasProperty('sshd.version') ? project.properties['sshd.version'] : "2.2.0"
 // commonIo has included by apache hadoop so we use the identical version
 versions["commonsIo"] = project.hasProperty('commonsIo.version') ? project.properties['commonsIo.version'] : "2.4"
+// commonIo has included by apache kafka so we use the identical version
```
saivirtue: typo? should we use guava here?

ohara-common/src/main/java/com/island/ohara/common/cache/RefreshableCache.java
```diff
@@ -0,0 +1,120 @@
...
+    private Cache<K, V> cache = null;
+    private Supplier<Map<K, V>> supplier = null;
+    private Duration frequency = Duration.ofSeconds(5);
+
+    private Builder() {}
+
+    public Builder<K, V> cache(Cache<K, V> cache) {
+      this.cache = Objects.requireNonNull(cache);
+      return this;
+    }
```
saivirtue: 看起來`RefreshableCache`是extend `Cache`的interface，在建立constructor時還需要傳入Cache物件嗎?

chia7712: yep, RefreshableCache compose of cache and a thread which keeps updating cache. Accepting a cache object can reduce the duplicate code. It means we don't need to accept the options same to Cache.Builder.

gradle/dependencies.gradle
chia7712: YOU ARE RIGHT :)

Issue #1087, opened 2019-05-09
Cannot run streamApp normally in pipeline


konekoya:
**Describe the bug**
After starting a basic streamApp in pipeline page. StreamApp will be listed in the k8s console but with a wired error: __Back-off restarting failed container__

**To Reproduce**
This can be reproduced on demo env

- Navigate to http://10.2.0.24:5050
- Create a worker and a new pipeline
- Create two new topics from Service -> Broker page
- Add these two topics into the pipeline you just created and add a new streamApp as well
- Add necessary config for StreamApp
- Start the StreamApp. It should have not problem starting the StreamApp but the error will be shown on k8s console

**Expected behavior**
There shouldn't be any error when starting a StreamApp

**Screenshots**

**Operation System (please complete the following information):**
I'm using demo env...

**Additional context**
Actual error from k8s' log

```
Exception in thread "main" java.lang.IllegalArgumentException: /tmp/streamApp-4917433142574234469.jar exists
	at com.island.ohara.common.util.CommonUtils.requireNotExist(CommonUtils.java:542)
	at com.island.ohara.common.util.CommonUtils.copyURLToFile(CommonUtils.java:602)
	at com.island.ohara.streams.StreamApp.findStreamAppEntry(StreamApp.java:173)
	at com.island.ohara.streams.StreamApp.lambda$main$4(StreamApp.java:142)
	at com.island.ohara.kafka.exception.CheckedExceptionUtils.wrap(CheckedExceptionUtils.java:45)
	at com.island.ohara.streams.StreamApp.main(StreamApp.java:137)
```

saivirtue:
Thanks @konekoya report this bug :)

PR #1088, opened 2019-05-09
#1087 Cannot run streamApp normally in pipeline

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1087)


chia7712:
Could you share the root cause? Incorrect path?

saivirtue:
> Could you share the root cause? Incorrect path?

See issue #1087, I have add the root cause in there

chia7712:
> This is due to the file copy function CommonUtils.copyURLToFile will check the destination file should not be exist, but streamApp will create a temp file first...

We should add a variety (or a flag) to copyURLToFile() to enable it to overwrite an existent file. That can help other people to handle the issue like this one.

saivirtue:
> > This is due to the file copy function CommonUtils.copyURLToFile will check the destination file should not be exist, but streamApp will create a temp file first...
> 
> We should add a variety (or a flag) to copyURLToFile() to enable it to overwrite an existent file. That can help other people to handle the issue like this one.

Well, the overwrite case  is rare, I think we can change the parameters to ((URL source, **String destinationFileName**, int connectionTimeout, int readTimeout) ; And add a small reminder in javadoc
how's that sound?

chia7712:
Another way is to add an new helper method which can download URL to a temporary file directly.
For example:

public static File downloadUrl(URL url)

The method iron out the temporary file issue.

saivirtue:
> public static File downloadUrl(URL url)

sounds great +1. file another issue instead.
BTW, this pr will be merge soon if there was no any other issue.

chia7712:
It would be better to have a test for this case...

saivirtue:
> It would be better to have a test for this case...

We have a streamApp test in `it` module, but there seems have some weird response of containerStatus from K8sClient (already tell to @jackyoh , reference issue is in #1091 ) ; this situation cause streamApp it test could run normally but actually is fail (CrashLoopBackOff)

chia7712:
We can’t verify the startup for s streamapp by UT?

saivirtue:
> We can’t verify the startup for s streamapp by UT?

as #1091 describe, streamApp will get `Running` even if the actual container(s) are `CrashLoopBackOff`

chia7712:
My point is - could we reproduce this bug in our testing code?Could we mock the incorrect information in UT?

saivirtue:
這邊的狀況是，錯誤的是"K8s 內Pod的containers"，而不是"Pod本身"；我們透過K8sClient互動的物件都是"Pod"
UT上無法mock K8s內部的狀態

saivirtue:
> we can assume that streamapp is able to download file

OK, now I got your point.
first, I can check the file download part, and this test case is my mistake :( , see issue #1101
but for this issue root cause, I think we can separate two different view :
1. from container view : root cause is download file fail
2. from ohara manager view : root cause is k8s client could not response correct status of container

In the second view, I think there is no way to add unit test, at least for now.

chia7712:
> from ohara manager view : root cause is k8s client could not response correct status of container

其實也可以做UT。因為k8s client是透過restful溝通，我們可以透過Akka fake一個k8s server並回傳#1091你列舉的json response，然後看看k8s client是否有轉成我們預期的結果

Issue #1135, opened 2019-05-13
Add POST restful api of streamApp property


saivirtue:
Since streamApp list page is independent from streamApp property page, we need add a `POST` method to create a streamApp component in pipeline.
Since the **List Page** and **Property Page** of streamApp should be separated, there are some API interface changes :

###  POST /v0/stream/jars
```diff
// Response
[
  {
+   "workerClusterName": "wk01",
    "id": "1b022c59-93f9-452c-a062-f8e4cb6c00fe",
-   "jarName": "ohara-streams.jar",
+   "name": "ohara-streams.jar",
    "lastModified": 1547141282866
  },
  {
+   "workerClusterName": "wk01",
    "id": "1a012c59-53de-3381-a062-f8e29f4c00fe",
-   "jarName": "other.jar",
+   "name": "other.jar",
    "lastModified": 1547141282889
  }
]
```
### GET /v0/stream/jars?cluster=wk01
```diff
// Response
[
  {
+   "workerClusterName": "wk01",
    "id": "1b022c59-93f9-452c-a062-f8e4cb6c00fe",
-   "jarName": "ohara-streams.jar",
+   "name": "ohara-streams.jar",
    "lastModified": 1547141282866
  },
  {
+   "workerClusterName": "wk01",
    "id": "1a012c59-53de-3381-a062-f8e29f4c00fe",
-   "jarName": "other.jar",
+   "name": "other.jar",
    "lastModified": 1547141282889
  }
]
```
### POST /v0/stream/property
```diff
// Request
{
  "jarId": "d23e7dfa52",
  *option* "name": "my-new-app",
  *option* "from": [
    "topic1"
  ],
  *option* "to": [
    "topic2"
  ],
  *option* "instances": 3
}
// Response
{
  "workerClusterName": "wk01",
  "id": "d312871a-4a05-488d-aae0-c8b27c5312c2",
  "name": "my-new-app",
  "instances": 1,
  "jarInfo": {
    "id": "1b022c59-93f9-452c-a062-f8e4cb6c00fe",
    "name": "new-name.jar",
    "size": 1234,
    "lastModified": 1542102595892
  },
  "from": [
    "topicA"
  ],
  "to": [
    "topicB"
  ],
  "lastModified": 1542102595892
}
```

required by #423

konekoya:
Is the __description__ also part of the response body? Should it be rendered/displayed in UI?

saivirtue:
[...]

PR #1152, opened 2019-05-14
#1135 Add POST restful api of streamApp property

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1135)


docs/rest_interface.md
```diff
@@ -2784,6 +2792,68 @@ Currently, this api is only used for changing jar name.
...
+
+*POST /v0/stream/property*
+
+**Example Request**
+
+1. jarId (**string**) — the used jar id
+1. name (**option(string)**) — new streamApp name ; default is "Untitled stream app"
+1. from (**option(array(string))**) — new source topics ; default is empty
+1. to (**option(array(string))**) — new target topics ; default is empty
+1. instances (**option(int)**) — new number of running streamApp ; default is 0
```
chia7712: Is **0** a valid value to us? If not, we can't use **0** as a default value. The default value should be a legal stuff.

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -310,10 +268,10 @@ private[configurator] object StreamRoute {
                           else
                             Future.successful(
                               previous.copy(
-                                name = req.name,
-                                instances = req.instances,
-                                from = req.from,
-                                to = req.to,
+                                name = if (req.name.isDefined) req.name.get else previous.name,
+                                instances = if (req.instances.isDefined) req.instances.get else previous.instances,
```
chia7712: getOrElse is more clear

saivirtue: :)

docs/rest_interface.md
saivirtue: make sense. edit the default value to 1

saivirtue:
retry

Issue #1114, opened 2019-05-10
Implement Pair and Tuple class to ohara


saivirtue:
Pair and tuple should have following attributes.

- Immutable
~- Comparable~
- Equal
- Hash code
~- Copy and replace~
- non-inheritable
- generic
- readable toString

related to #1105

PR #1156, opened 2019-05-14
#1114 Implement Pair class to ohara

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1114)


ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
```diff
@@ -0,0 +1,99 @@
...
+public final class Pair<L, R> implements Serializable {
+
+  // Serialization version
+  private static final long serialVersionUID = 1L;
+
+  private final L left;
+  private final R right;
+
+  private Pair(final L left, final R right) {
+    this.left = left;
```
chia7712: Should we avoid null?

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
```diff
@@ -0,0 +1,99 @@
...
+  public static <L, R> Pair<L, R> of(final L left, final R right) {
+    return new Pair<>(left, right);
+  }
+
+  /**
+   * get the left element from this pair
+   *
+   * @return left element
+   */
+  public L getLeft() {
```
chia7712: Could we address scala style. left() and right()

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * update the elements of pair. Will create a new immutable pair
+   *
+   * @param left new left element
+   * @param right new right element
+   * @param <NL> left element type
+   * @param <NR> right element type
+   * @return new immutable pair
+   */
+  public <NL, NR> Pair<NL, NR> update(final NL left, final NR right) {
```
chia7712: this method is weird...It just create an new pair which is not associated to origin one.

We should have two **copy** method for Pair. For example.

```java
  public <NR> Pair<L, NR> replaceRight(NR newRight) {
    return new Pair<>(left, newRight);
  } 
```

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
```diff
@@ -0,0 +1,99 @@
...
+  private final R right;
+
+  private Pair(final L left, final R right) {
+    this.left = left;
+    this.right = right;
+  }
+
+  /** Create an immutable pair of nulls */
+  @SuppressWarnings("rawtypes")
+  public static final Pair NULL = Pair.of(null, null);
```
chia7712: Is it necessary??

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
saivirtue: OK

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
saivirtue: In this case , I think user can use
```java
pair.update(pair.getLeft(), new_right)
```

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
saivirtue: maybe "empty pair" is less useful...I will remove the relative code

ohara-common/src/main/java/com/island/ohara/common/data/Pair.java
chia7712: It is almost equal to new Pair(pair.getLeft, new_right), right?

We can just remove this helper method if it does not give a bit benefit.

saivirtue: Sure. Update an immutable object is a wield operation...
I will remove this method

Issue #185, opened 2019-02-15
Support to use RocksDB as default store for Configurator


chia7712:
Configurator, currently, uses in-memory store to save objects. It is not persistent so we have to clone whole Configurator container to keep data. The clone is a kind of "snapshot" so there is likelihood that data loss.

- [x] Introduce RocksDB Store (#1166)
- [x] document usage of RocksDB (#1207)

vitojeng:
an important part for ohara backend.
I'm +1 on this issue.

PR #1166, opened 2019-05-14
#185 Introduce RocksDB Store

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #185)


ohara-configurator/src/main/scala/com/island/ohara/configurator/Configurator.scala
```diff
@@ -157,12 +139,17 @@ class Configurator private[configurator] (advertisedHostname: Option[String],
...
-      .frequency(cacheRefresh)
+      .frequency(cacheTimeout)
       .build()
   }
 
+  implicit val urlGenerator: UrlGenerator = new UrlGenerator {
+    override def url(id: String)(implicit executionContext: ExecutionContext): Future[URL] =
+      Future.successful(new URL(s"http://$hostname:$port/${ConfiguratorApiInfo.V0}/${JarsRoute.pathToJar(id)}"))
+  }
+
```
saivirtue: Does this mean `user cannot explicitly access jar route` ?

saivirtue:
I think you miss the **Assignee* * part 😆

chia7712:
> think you miss the *Assignee * part 😆

Done :)

docs/user_guide.md
```diff
@@ -79,6 +79,7 @@ All images are list below.
 docker run --rm -p ${port}:${port} --add-host ${nodeHostName}:${nodeHostIP} oharastream/configurator:0.5-SNAPSHOT --port ${port} --hostname ${host} --node ${SshUserName}:${SshPassword}@${NodeHostName}:${SshPort}
 ```
 
+- folder: the folder used to store data (default is random). Mount the volume if you want to keep your data after restarting Configurator
```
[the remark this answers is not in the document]
chia7712: The default folder is always on the tmp folder :)

ohara-configurator/src/main/scala/com/island/ohara/configurator/Configurator.scala
chia7712: Yep. We don’t encourage ohara user to download jar from Configurator.  See [docs](https://github.com/oharastream/ohara/blob/master/docs/rest_interface.md#jars)

ohara-configurator/src/main/scala/com/island/ohara/configurator/store/DataStore.scala
```diff
@@ -73,9 +73,9 @@ object DataStore {
       store.value(id).filter(classTag[T].runtimeClass.isInstance).map(_.asInstanceOf[T])
 
     override def values[T <: Data: ClassTag](implicit executor: ExecutionContext): Future[Seq[T]] =
-      store.values().map(_.values.filter(classTag[T].runtimeClass.isInstance).map(_.asInstanceOf[T]).toSeq)
+      store.values.map(_.values.filter(classTag[T].runtimeClass.isInstance).map(_.asInstanceOf[T]).toSeq)
```
vitojeng: nit:
**store.values()** would be more readable and maintainable in general naming conversions.

chia7712:
retry

ohara-configurator/src/main/scala/com/island/ohara/configurator/store/DataStore.scala
chia7712: ok

chia7712: The method accepts a implicit input - ExecutionContext - so I have it be paraemeterless style. It is similar to [Seq.sorted](https://www.scala-lang.org/api/2.12.8/scala/collection/Seq.html#sorted[B%3E:A](implicitord:scala.math.Ordering[B]):Repr)

vitojeng: Oh...It's my bad. 😢 
I didn't notice the `values()` method signature in the `Store` changed.

For implicit input, I would suggest keeping `()`, for example:
```scala
trait Store[K, V] extends ReleaseOnce {
  def values()(implicit executor: ExecutionContext): Future[Map[K, V]]
}
```

IMO, one of the naming conventions of Scala is the `def` method without parentheses usually means not change any state.

chia7712: > means not change any state.

We are on the same page. The method **values** does not cause any changes and side-effect to the instance at all. Personally, it should be ok to keep it be parameter-less.

vitojeng: DataStore access external storage, it would possibly cause some exception(side-effect).

chia7712: >DataStore access external storage, it would possibly cause exception(side-effect).

that is another complicated story about how pure of a function should be. At any rate, I'm ok to get rid of the parameterless from this method.

chia7712:
retry configurator

chia7712:
retry connector

Issue #1089, opened 2019-05-09
Add CommonUtils.downloadUrl function


saivirtue:
Since download a url to a local file is generally useful, we could use function interface to be more straight :

**download url, and return the local download file**
`public static File downloadUrl(URL url)`

See #1087

PR #1170, opened 2019-05-15
#1089 Add CommonUtils.downloadUrl function

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1089)


ohara-common/src/main/java/com/island/ohara/common/util/CommonUtils.java
```diff
@@ -598,6 +598,24 @@ public static void copyURLToFile(
     }
   }
 
+  /**
+   * download file from url
+   *
+   * @param url where the file is download from
+   * @return the downloaded file
+   */
+  public static File downloadUrl(URL url) {
```
chia7712: It would be better to have a variety accepting the timeout :)

saivirtue: sure :)

ohara-common/src/main/java/com/island/ohara/common/util/CommonUtils.java
```diff
@@ -598,6 +598,23 @@ public static void copyURLToFile(
...
 
+  /**
+   * download file from url
+   *
+   * @param url where the file is download from
+   * @param connectionTimeout the connection timeout
+   * @param readTimeout the read timeout
+   * @return the downloaded file
+   */
+  public static File downloadUrl(URL url, int connectionTimeout, int readTimeout) {
```
chia7712: Could we use java.time.Duration to replace the **primitive type**??? Duration is more readable.

saivirtue: Hmm...let me try it.

saivirtue:
retry

ohara-streams/src/main/java/com/island/ohara/streams/StreamApp.java
```diff
@@ -164,15 +166,7 @@ public static void main(String[] args) {
...
-    // create a tempFolder and new a file instance : /tmp/streamApp-XXXXX/streamApp.jar
-    File tmpFolder = CommonUtils.createTempFolder("streamApp-");
-    File outputFile = new File(tmpFolder, "streamApp.jar");
-
-    URL url = new URL(jarUrl);
-
-    // Download the jar
-    CommonUtils.copyURLToFile(url, outputFile, CONNECT_TIMEOUT, READ_TIMEOUT);
-    return outputFile;
+    return CommonUtils.downloadUrl(new URL(jarUrl), CONNECT_TIMEOUT, READ_TIMEOUT);
```
chia7712: Does compiler accept this code?

saivirtue: :(

Issue #904, opened 2019-04-24
Refactor crane to collie


saivirtue:
To simplify the "cluster concepts" in ohara, any cluster implement should extends from collie.
We have two different cluster group by now :
1. service cluster : zk, bk, wk  => use collie
2. container cluster : stream  => use crane

Since `crane` is more like "cluster manager", and `warehouse` is more like `clusters`, it is fit the structure of collie.

This issue addresses to the following tasks:
- delete Crane
- move StreamWarehouse -> StreamCollie

saivirtue:
done this huge refactor...

PR #1212, opened 2019-05-20
 #904 Refactor crane to collie

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #904)


ohara-agent/src/main/scala/com/island/ohara/agent/ContainerCollie.scala
```diff
@@ -122,6 +124,7 @@ object ContainerCollie {
   val ZK_SERVICE_NAME: String = "zk"
   val BK_SERVICE_NAME: String = "bk"
   val WK_SERVICE_NAME: String = "wk"
+  val STREAM_SERVICE_NAME: String = "stream"
```
chia7712: Given that limit of container name, could we have a shorter one?

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
```diff
@@ -14,41 +14,63 @@
...
+    Future.successful(
+      StreamClusterInfo(
+        name = clusterName,
+        imageName = first.imageName,
+        nodeNames = containers.map(_.nodeName),
+        state = {
+          // we only have two possible results here:
+          // 1. only assume cluster is "running" if all of containers are running
+          // 2. the cluster state is always "dead" if one of container state was not running
+          Some(containers.map(_.state).reduce[String] {
```
chia7712: containers.map(_.state).exist(ContainerState.DEAD.name) is more suitable for this case since 
1. faster (it doesn't need to traverse all containers)
1. more readable -> if (dead exist)  ContainerState.DEAD.name else ContainerState.RUNNING.name

BTW, how we handle the dead streamapp???

saivirtue: > containers.map(_.state).exist(ContainerState.DEAD.name) 

Consider the same logic (iff all containers are running), this will be 
`val dead = !containers.forall(_.state == ContainerState.RUNNING.name)`
If this is more readable (compare to **reduce** ), I can rewrite this part.

> how we handle the dead streamapp

Only one way, stop (remove cluster) and start it.

ohara-agent/src/main/scala/com/island/ohara/agent/ContainerCollie.scala
saivirtue: Do we encounter the limit?
Since the container name for streamApp is totally control by `streamRoute`, less word less meaning, I prefer use **stream** here...
(**st**, **sm** seems strange here 😂 )

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
chia7712: > val dead = !containers.forall(_.state == ContainerState.RUNNING.name)

如果可以的話盡量別用"否定"判斷 因為我們人腦在閱讀這類的敘述的時候很容易打結(例如公投題目...)

> Only one way, stop (remove cluster) and start it.

我疑惑的是一個stream cluster的部分死亡真的要當成"叢集死亡"嗎??

ohara-agent/src/main/scala/com/island/ohara/agent/ContainerCollie.scala
chia7712: >(st, sm seems strange here

agreed :)  Could you document it?

saivirtue: OK

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
saivirtue: > 如果可以的話盡量別用"否定"判斷

OK, I will use "positive expression" instead

> 真的要當成"叢集死亡"嗎??

We cannot decide the recover steps after some containers were not running
 - k8s will auto retry fail container for limit times
 - kafka streams will auto balance consuming instances after some containers were fail

If the fail container still dead, restart the whole cluster is more easy for two reasons:
1. start streamApp containers are fast
2. we can solve the fail exception (ex, broker down) if restart the cluster was work

saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
chia7712: > We cannot decide the recover steps after some containers were not running

My point was not recovery. A part of dead containers should **not** equal to a dead streamapp cluster. It still work since kafka auto-move the loading from dead consumer to alive one. 

> restart the whole cluster 

I don't prefer to restart **all** streamapp containers since it burn out our availability. The restart produce a true **downtime** to our system.

Let us imagine that a container node is gone and all streamapp clusters related to this node are going to die also. Does it make sense?

chia7712: BTW, we can move this discussion to another issue. It is ok to keep current behavior before we reach the consensus.

saivirtue: 我想這邊我可能沒有說的很清楚
> It still work since kafka auto-move the loading from dead consumer to alive one.

這個是streamApp運行時的結果，但這裡的`state`是指使用者看到stream cluster的狀態，即使後面kafka會幫我們做很多事情，我們依舊需要把cluster的狀態反映給使用者 (部分失敗意味著還是有問題，我想這個可以認同?)；需要recover container的工作，看未來有一個類似container management的介面來做這一塊會不會好一點?

chia7712: >  (部分失敗意味著還是有問題，我想這個可以認同?)

A cluster means it is doable to tolerant a part of nodes which are dead. For example, we don't say a hadoop cluster is dead even if one of datanodes is gone. I feel we should have the same **concept** if we have chosen the **cluster** naming.

> 我們依舊需要把cluster的狀態反映給使用者

This is another issue about how to tell user that a part of nodes (containers) are gone. For instance, we can add a field - deadNodes - to interface ClusterInfo.

> 需要recover container的工作，看未來有一個類似container management的介面來做這一塊會不會好一點?

this is another important issue. How ohara handle the auto-recovery....

chia7712: Speak clearly, we say a cluster is dead only if we ensure that the cluster does not work anymore.

saivirtue: Let's move the discussion here XD
#1220

saivirtue:
retry it

saivirtue:
retry it

chia7712:
@saivirtue Please clear the orphan containers from all nodes if you encounter the timeout.

saivirtue:
retry it

saivirtue:
The failed it tests : testMultiBkClustersOnSingleNode, testMultiWkClustersOnSingleNode are not related this PR, will merge this PR later.

chia7712:
+1

Issue #968, opened 2019-04-30
Refactor stream route to be more tidy


saivirtue:
Currently, the `RouteUtils` contains a lot of useful common APIs for most router using, we need to apply this to stream router.
- move `jmxPort` to `port` in order to use the benefit of port conflict check
- use some basic cluster checks of RoteUtils
- refactor property api of streamApi (use `RouteUtils.basicRoute`)

PR #1296, opened 2019-05-27
#968 Refactor stream route to be more tidy

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #968)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -299,57 +276,73 @@ private[configurator] object StreamRoute {
...
+                                      .cluster(formatUniqueName(data.id))
+                                      .filter(_._1.isInstanceOf[StreamClusterInfo])
+                                      .map(_._1.asInstanceOf[StreamClusterInfo])
+                                  } else {
+                                    clusterCollie
+                                      .streamCollie()
+                                      .creator()
+                                      .clusterName(formatUniqueName(data.id))
+                                      .instances(data.instances)
+                                      .imageName(IMAGE_NAME_DEFAULT)
```
chia7712: Why not using user-defined image?

saivirtue: stream cluster is more like an application, the image name is less usage for them.

saivirtue:
rety

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
chia7712: Is it a bug if user send StreamClusterCreationRequest with a specific image name?

saivirtue:
retry

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
saivirtue: User cannot send a **StreamClusterCreationRequest** to create a stream cluster...The StreamClusterCreationRequest object is used for **extend ClusterCreationRequest** and then do the cluster check. In the stream route, I use another request body for user (StreamPropertyRequest) to define a streamApp property.

Maybe this is a little weird...since the other cluster creation uses **POST**, the stream cluster creation is auto created in start command instead.

chia7712: Got it.

saivirtue:
retry configurator

Issue #1302, opened 2019-05-28
Increase file size limit in JarRoute and StreamRoute


saivirtue:
From [document](https://doc.akka.io/docs/akka-http/current/configuration.html):
```diff
# Default maximum content length which should not be exceeded by incoming request entities.
# Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
# Note that it is not necessarily a problem to set this to a high value as all stream operations
# are always properly backpressured.
# Nevertheless you might want to apply some limit in order to prevent a single client from consuming
# an excessive amount of server resources.
#
# Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
# programmatically via `withSizeLimit`.)
max-content-length = 8m

# The maximum number of bytes to allow when reading the entire entity into memory with `toStrict`
# (which is used by the `toStrictEntity` and `extractStrictEntity` directives)
max-to-strict-bytes = 8m
```

Since streamApp use `toStrict` to solve the [issue](https://github.com/akka/akka-http/issues/1216#issuecomment-311973943), we need to increase the file limit configuration to a suitable value, says **50m** by programmatically use `withSizeLimit`.
This issue will add the limit into:
- JarRoute upload api
- StreamRoute upload api

The StreamRoute upload api will be refactor in #1252

chia7712:
Are you talking about file size? If so, I recommend to change only the jar route instead of global configs.

For example, we can compose the file upload directive by withSizeLimit

saivirtue:
> For example, we can compose the file upload directive by withSizeLimit

As far as I know, the default `max-to-strict-bytes` is check before `withSizeLimit` ; User will face an exception like **Request Content-Length XXXX exceeds the configured limit of 8388608**.
I think the cause is from global config since I don't set any limit in stream route

saivirtue:
BTW, the error is thrown by streamApp upload api instead of jarRoute...which inspired me the error is from `toStrict` part...

vitojeng:
+1

The user only can run streamapp with fat-jar at this time, so we really need to increase the file size limitation when upload streamapp jar.

chia7712:
> As far as I know, the default max-to-strict-bytes is check before withSizeLimit

Not really. please read the [docs](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/misc-directives/withSizeLimit.html) 

```
. Limit given as parameter overrides limit configured with akka.http.parsing.max-content-length.
```

[...]

PR #1310, opened 2019-05-28
#1302 Increase file size limit in JarRoute and StreamRoute

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1302)


ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestJarRoute.scala
```diff
@@ -44,33 +43,48 @@ class TestJarRoute extends SmallTest with Matchers {
...
-    Await.result(jarApi.list, 30 seconds).size shouldBe 1
+    result(jarApi.list).size shouldBe 1
+
+    f.deleteOnExit()
+  }
+
+  @Test
+  def testUploadOutOfLimitFile(): Unit = {
+    val bytes = new Array[Byte](RouteUtils.DEFAULT_JAR_SIZE_BYTES.toInt + 1)
+    Random.nextBytes(bytes)
```
chia7712: the default value - 0 - is ok in this case so we don't need to initialize array manually.

saivirtue: I'm not sure get the point. Do you mean use `val bytes = new Array[Byte](0)` instead ?

chia7712: We don't need to assign random **content** to the byte array since the default value of each byte is 0.

saivirtue: Got it

Issue #1350, opened 2019-05-31
Cannot remove exited containers of a cluster in docker environment


saivirtue:
In docker environment, we can remove a container "gracefully" by calling the following commands sequentially :
```
docker stop XXXX
docker rm XXX
```
By default, `docker stop` will send another **kill** signal after stop command is not complete for over 10 seconds. We need to control suitable timeout between docker and ssh connection to avoid possible situation that ssh connection closed but docker stop command is not complete.

Another problem is `ClusterCache` only get **active** containers, we should handle the **exited** containers in docker remove command.

Following tasks will be accomplish:
- change the default `ClusterCache` will get only **running** containers in docker and k8s env
- make sure all collies could only get **running** containers in ssh env
- make sure streamApp should assume cluster **dead** only if all containers are dead
- make sure **exited** containers could be removed during cluster removing

saivirtue:
/cc @chia7712 
I have found this problem in testing my streamApp cluster...If there was no other concern, I will fix this bug today.

chia7712:
Nice finding. +1

chia7712:
BTW, does it cause error if we try to **stop** a exited container?

saivirtue:
> BTW, does it cause error if we try to **stop** a exited container?

Seems not. a exited container is actually **stop** working. Calling stop command will do nothing.

chia7712:
Pardon me, is there a bug in this issue?

saivirtue:
> Pardon me, is there a bug in this issue?

In my IT tests, if I ran a streamApp cluster and it failed later, use the `cluster.remove()` command will remove the clusterCache successfully, but the actual container are not removed.

chia7712:
> In my IT tests, if I ran a streamApp cluster and it failed later, use the cluster.remove() command will remove the clusterCache successfully, but the actual container are not removed.

The root cause is that ssh collie does care for **active** containers only. Hence, the **exited** containers disappear from the cluster list.

saivirtue:
> The root cause is that ssh collie does care for **active** containers only. Hence, the **exited** containers disappear from the cluster list.

True. ClusterCache will get all **active** containers only, but since I use cluster remove function, the clusterCache remove the cluster (either running or exited) is right behavior, but docker should take the responsibility to remove all containers that belong to this cluster.

chia7712:
這個議題應該要考慮兩個點：
[...]

PR #1351, opened 2019-05-31
#1350 Cannot remove exited containers of a cluster in docker environment

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1350)


ohara-agent/src/main/scala/com/island/ohara/agent/ssh/BasicCollieImpl.scala
```diff
@@ -68,15 +68,15 @@ private abstract class BasicCollieImpl[T <: ClusterInfo: ClassTag, Creator <: Cl
...
-                                 if (force) client.forceRemove(containerInfo.name)
-                                 else {
-                                   client.stop(containerInfo.name)
-                                   client.remove(containerInfo.name)
-                               }))
+          .map(node => {
+            if (force) {
+              dockerCache.exec(node, client => client.forceRemove(containerInfo.name))
+            } else {
+              // We make sure container removed after stopped command, use two ssh connections to execute
```
chia7712: Do we remove the container too quick? Sending two connection can resolve this issue? why?

saivirtue: Yes, it is because we remove container too quick.
I think separate two `DockerClient` commands could reduce the intensive calling.
Have good ideas about this?

chia7712: We can just increase the wait time of docker stop.

ohara-agent/src/main/scala/com/island/ohara/agent/Collie.scala
```diff
@@ -72,7 +73,20 @@ trait Collie[T <: ClusterInfo, Creator <: ClusterCreator[T]] {
...
+  /**
+    * fetch all clusters and belonging containers from cache
+    * this function will only get running containers
+    *
+    * @param executionContext execution context
+    * @return cluster and containers information
+    */
+  def clusters(implicit executionContext: ExecutionContext): Future[Map[T, Seq[ContainerInfo]]] =
+    clusterWithAllContainers.map(entry =>
+      entry.map { case (info, containers) => info -> containers.filter(_.state == ContainerState.RUNNING.name) })
```
chia7712: Could you document it ? For example: Currently, both k8s and pure docker have the same context of **RUNNING** so it is ok to filter container via **RUNNING**.

ohara-agent/src/main/scala/com/island/ohara/agent/Collie.scala
```diff
@@ -72,7 +73,20 @@ trait Collie[T <: ClusterInfo, Creator <: ClusterCreator[T]] {
...
+
+  // for sub-class use only,
+  protected def clusterWithAllContainers(
```
chia7712: document the reason of this specific method

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
```diff
@@ -33,19 +33,24 @@ import scala.concurrent.{ExecutionContext, Future}
...
     Future.successful(
       StreamClusterInfo(
         name = clusterName,
         imageName = first.imageName,
         nodeNames = containers.map(_.nodeName),
         // In StreamApp, we only use jmx port for exposing
-        jmxPort = first.portMappings.head.portPairs.head.containerPort,
+        jmxPort =
+          //
+          if (first.portMappings.isEmpty) -1
```
chia7712: Is jmx port not open in 0.5+ ?

ohara-agent/src/main/scala/com/island/ohara/agent/Collie.scala
saivirtue: OK

ohara-agent/src/main/scala/com/island/ohara/agent/Collie.scala
saivirtue: sure

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
saivirtue: jmx port is fetched by the expose ports of container ; we cannot get the "actual" jmx port if all containers were exited.
Let me address some comment here :)

chia7712: In ssh mode, it is doable to fetch (implemented by ```docker inspect```)  the envs from a exited container.

@jackyoh does k8s have same behavior?fetch envs from an exited container?

saivirtue: k8s should OK to do the same job. Let me add an issue to accomplish this.
#1378
I think @jackyoh is too busy in this milestone cycle...XD

chia7712: @saivirtue DockerClient.containers is already able to return all containers (even dead) with **all** envs since it is using both  ```docker ps``` and ```docker inspect```.

It seems to me the porting mapping should NOT be empty and we should directly throw exception for empty mapping.

saivirtue: > It seems to me the porting mapping should NOT be empty and we should directly throw exception for empty mapping.

Dead container would not expose port, even we run the container with explicit port mapping.
But you are right, this line imply **port is empty, should throw exception**, I will change the fetching rule for jmx port

chia7712: > Dead container would not expose port, even we run the container with explicit port mapping.

We **can** inspect an exited container even though the docker container is exited. Hence, the porting mapping must be **not** empty if we had exposed port on the (exited) container. That is to say, the empty porting mapping is an exception than a expected behavior.

saivirtue: > We can inspect an exited container even though the docker container is exited.

This is not the current behavior, should I fire an issue to accomplish this?

saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
chia7712: bq. This is not the current behavior

[link](https://github.com/oharastream/ohara/blob/master/ohara-agent/src/main/scala/com/island/ohara/agent/docker/DockerClientImpl.scala#L226).

saivirtue: Oh, I should explain more clearly :(
As far as I know, `DockerClient` use `docker inspect` to get env and hostname, but the expose port was fetched by `docker ps`, which I think maybe we can change the behavior as same as env did.

chia7712: > As far as I know, DockerClient use docker inspect to get env and hostname, but the expose port was fetched by docker ps, which I think maybe we can change the behavior as same as env did.

Maybe not. We should keep docker client pure :) Auto-fill something callers don't expected may be dangerous. At any rate, the latest code look good to me.

saivirtue:
retry agent

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry client

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
finally...all passed. Close this pr later.

Issue #1252, opened 2019-05-22
Migrate upload jar api of streamApp to JarRoute api


saivirtue:
StreamApp upload api support:
- upload multiple jars
- delete specific jar
~- rename specific jar~ (question in #1426)

And the above APIs are similar with JarRoute api. This issue is intend to ease the duplicate work of jar operations.

**We need to finish the following tasks**
- [x] check if there is any issue for the migration work
- [x] add the same operation to JarRoute (#1427)
-- support `group` in jarRoute
- [x] apply same rule from StreamRoute jar api to JarRoute (#1443)
- [x] move the used api from /stream/jar to /jar in manager (move to #1560)
- [x] remove the stale stream api (move to #1560)

chia7712:
Nice one. Is this issue too big to complete in 0.5 ? If not, please remove umbrella label and attach 0.5 label.

saivirtue:
This umbrella issue require @konekoya to accomplish together. Will check the available time of him.
And also, this issue has multiple tasks, as we discuss the umbrella usage, I think the umbrella label should be fine. `v0.5.0` label will be added after check.

chia7712:
> this issue has multiple tasks, as we discuss the umbrella usage

An umbrella should be used to include multiples issues **and** all of them can be resolved in a version release. Filing many **trivial** issue with **nothing** description is not good to us, and it is not convenient to trace story from one issue to another issue. Especially, all stories are record in root one...

chia7712:
> This umbrella issue require @konekoya to accomplish together.

And it would be better to update APIs without **breaking** the legacy one. We had updated many APIs with this way, and you can file a issue to trace the stale APIs. (for example #518). As a backend developer, we can't always wait/block frontend developer...

saivirtue:
> Is this issue too big to complete in 0.5 ? If not, please remove umbrella label

> An umbrella should be used to include multiples issues and all of them can be resolved in a version release. 

Are these two sentences contradict each other?

This issue is just want to change the api without any harmful side-effect:
1. add the same operation to JarRoute api (**preserve old function**)
2. move the used api from /stream/jar to /jar in manager (**move old api to new api**)
3. remove the stale stream api (**clean old api**)

at any rate, task 1 could be implemented without waiting

chia7712:
> Are these two sentences contradict each other?

" all of them can be resolved " -> " all of them can't be resolved "

PR #1443, opened 2019-06-12
#1252 Apply same rule from streamRoute to jarRoute

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1252)


ohara-configurator/src/main/scala/com/island/ohara/configurator/jar/JarStore.scala
```diff
@@ -78,12 +78,18 @@ trait JarStore extends Releasable {
...
   }
 
-  def jarInfos()(implicit executionContext: ExecutionContext): Future[Seq[JarInfo]]
+  /**
+    * retrive all jars in jar store
+    * @param group the filter group name
+    * @param executionContext execution context
+    * @return jars description
+    */
+  def jarInfos(group: Option[String])(implicit executionContext: ExecutionContext): Future[Seq[JarInfo]]
```
chia7712: The **true-or-false** argument should be spitted to two methods.
1. jarInfos() - no filter. It means all jars should return
1. jarInfos(group: String) - retrieve the jars for input group

saivirtue: ok

saivirtue:
> We should avoid to encourage caller to use **none** as input arguments.

Let me refactor some methods in **JarStore**

Issue #1281, opened 2019-05-24
Replace create-react-app's NODE_PATH settings with a jsconfig.json


konekoya:
`NODE_PATH` is deprecated (__A deprecated message is displayed when starting the client dev server__) and will be completed removed in a future major release. We should use the new recommended setup -> create a new `jsconfig.json` in the project root (I'm assuming this is in the create-react-app root), and use the settings from the [docs](https://facebook.github.io/create-react-app/docs/importing-a-component#absolute-imports)

This seems like a fairly easy upgrade. But As we're using several `NODE_PATH`. The config from the docs is not working properly in our App. Hopefully, we can figure this out before the next major release.

/cc @wu87988622

konekoya:
Although the multiple absolute paths are not supported yet. I decide to upgrade and use the new recommended approach (use a `jsconfig.js`) anyway as the `NODE_PATH` could be removed any time from a major create-react-app release

PR #1468, opened 2019-06-16
#1281 Replace create-react-app's NODE_PATH settings with a jsconfig.json

konekoya:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1281)


konekoya:
@wu87988622 due to the restriction of create-react-app 3.1.0 We no longer to support absolute path for common and page dirs.

We were able to do this before:
```js
import module from common/moduleDir
import {module} from common/moduleDir
``` 

But now:

```js
import module from components/common/moduleDir
import {module} from components/common/moduleDir
```

So basically, the absolute path now only support the first level right under the `src` dir. So we are still able to do:

```js
import module from api/module
import aConstant from constants/file
impor utils from utils/myUtils
// So on and so forth
```

But the benefit we gain here is the ability to use the goto

konekoya:
retry manager

konekoya:
retry manager

konekoya:
Unrelated fail case, merge the PR anyway

Issue #1477, opened 2019-06-17
Make sure all in-flight requests are closed


chia7712:
this is another cause of jvm crash when we are running tests for configurator module.

PR #1485, opened 2019-06-18
#1477 Make sure all in-flight requests are closed

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1477)
loop this PR 10 times, and all pass


ohara-configurator/src/main/scala/com/island/ohara/configurator/Configurator.scala
```diff
@@ -63,7 +63,7 @@ class Configurator private[configurator] (val hostname: String, val port: Int)(i
     extends ReleaseOnce
     with SprayJsonSupport {
 
-  private[this] val initializationTimeout = 10 seconds
+  private[this] val initializationTimeout = 30 seconds
```
vitojeng: Do we really need to increase the timeout value by three times? 😆

chia7712: the increased timeout is used to **handle** the in-flight requests. We have to send "50x" to all requests.

vitojeng: Got it. That's great.

Issue #1466, opened 2019-06-16
Add more arguments check for Agent and DockerClient


chia7712:
We all need more tests!

jackyoh:
+1

vitojeng:
+100

Issue #1488, opened 2019-06-18
Update the docs URL in the API response body


konekoya:
As our docs have moved. The docs URL from the RESTful API response body should be updated as well. The current API URL will point to a 404 page from GitHub

Res body:

```json
{
    "code": "Unsupported API: v0/containers",
    "message": "please see link to find the available APIs",
    "stack": "N/A",
    "apiUrl": "https://github.com/oharastream/ohara/blob/master/docs/rest_interface.md"
}
```

konekoya:
I will file a PR to fix this if I get a chance today

konekoya:
@vitojeng A quick question. Is this (👇 ) the correct URL of our new RESTful API home?
https://oharastream.readthedocs.io/en/latest/rest_interface.html

konekoya:
PR is up https://github.com/oharastream/ohara/pull/1489

chia7712:
@konekoya This issue break the [test](https://builds.is-land.com.tw/job/PostCommit-OHARA/428/testReport/). Could you file a issue for the flaky?

konekoya:
> @konekoya This issue break the test. Could you file a issue for the flaky?

Oops, thought It was only an easy fix... Will give it a look as soon as I can

Issue #1417, opened 2019-06-07
Gradle error happens on testing configurator


chia7712:
[link](https://builds.is-land.com.tw/job/PreCommit-OHARA/2617/console)

```
18:06:37 * What went wrong:
18:06:37 Execution failed for task ':****-configurator:test'.
18:06:37 > Process 'Gradle Test Executor 128' finished with non-zero exit value 1
18:06:37   This problem might be caused by incorrect test process configuration.
18:06:37   Please refer to the test execution section in the User Manual at https://docs.gradle.org/5.4.1/userguide/java_testing.html#sec:test_execution
18:06:37 
18:06:37 * Try:
18:06:37 Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
18:06:37 
18:06:37 * Get more help at https://help.gradle.org
```

In testing we make a resource leak of rocksdb so the jvm invoked by Junit can't terminate successfully.

Maybe related to #1435

chia7712:
This bug happens on testing only so I don't backport it to 0.5. FYI @eechih

eechih:
> This bug happens on testing only so I don't backport it to 0.5.

OK

chia7712:
已經一段時間沒看過此錯誤了 所以關閉議題

PR #1501, opened 2019-06-18
#1466 Add more arguments check for Agent and DockerClient

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1466)


ohara-common/src/main/java/com/island/ohara/common/Builder.java
```diff
@@ -0,0 +1,33 @@
...
+ * @param <T> type of final result
+ */
+public interface Builder<T> {
+  /**
+   * produce a instance. This is the final operation of chain of this builder. DON'T reuse this
+   * builder since all inner objects are undefined after instantiating a object.
+   *
+   * @return an object
+   */
+  T result();
```
saivirtue: I think `build()` is more straightforward since we are create a "builder"

chia7712: ok

chia7712:
Configurator error is related to #1417 and #1488.

Issue #1674, opened 2019-07-01
the generated default value from JsonRefiner is constant than random


chia7712:
JsonRefiner offers many friendly and useful function in parsing json request. The random default value play a important rule in simplifying the request format since it allows user to ignore many fields from request. However, the random default value is not true random since the **value** is decided in creating. It means the later request see the identical value than a true random stuff :( 

- [x] for 0.7 (#1675)
- [x] for 0.6 (#1677)

PR #1675, opened 2019-07-01
#1674 the generated default value from JsonRefiner is constant than r…

chia7712:
…andom

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1674)


ohara-client/src/test/scala/com/island/ohara/client/configurator/v0/TestWorkerApi.scala
```diff
@@ -316,6 +316,15 @@ class TestWorkerApi extends SmallTest with Matchers {
...
     creation.jarKeys.size shouldBe 0
+
+    val creation2 = WorkerApi.WORKER_CREATION_JSON_FORMAT.read(s"""
+                                                                 |  {
+                                                                 |    "name": "$name",
+                                                                 |    "nodeNames": ["$nodeName"]
+                                                                 |  }
+                                                                     """.stripMargin.parseJson)
+    creation2.clientPort should not be creation.clientPort
+    creation2.jmxPort should not be creation.jmxPort
```
saivirtue: could we add an assert to verify the different creation should have different name ?

chia7712: this test is of paring json string so the value of name is what we input via string. Hence, the identical name is ok in parsing.

Issue #810, opened 2019-04-18
Enhance streamApp to support custom setting


saivirtue:
like #480 did, we need to make streamApp more flexible in settings.

This issue has the following tasks:
- [x] Support custom configs for streamApp (#1726)
-  Change the interface of `start()` to accept parameters (**OStream**, **ConfigDef**)
-  Config is like `SettingDefinition` (should extend JSON serialize / deserialize)
-  ConfigDef should manage the Config class
-  Add new abstract method `ConfigDef config()` for user to define configurations
- [x] Support custom config for stream API (#1929)
- [x] change prefix from stream to streams (move to issue #2115)

PR #1726, opened 2019-07-04
 #810 Support custom configs for streamApp

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #810)


gradle/dependencies.gradle
```diff
@@ -27,6 +27,11 @@ versions["scala"] = project.hasProperty('scala.version') ? project.properties['s
 if (versions["scala"] == "2.12") versions["scala"] = "2.12.8"
 if (versions["scala"].startsWith("2.12")) versions["scalaMinor"] = "2.12"
 else throw new GradleException("${versions["scala"]} is not supported. ohara requires scala 2.12")
+// define dependencies version
+if(versions["kafka"] == "2.2.1") {
+  versions["jackson"] = "2.9.8"
+  versions["reflections"] = "0.9.11"
+}
```
chia7712: seems the flexible to version is too hard to maintain. Perhaps we should disable the changeable version of deps.

saivirtue: I am not intend to expose the `reflection class` to user, and it should not too.
The whole idea is to open a door for user to define custom configuration by method `ConfigDef config()`. Since I'm not sure the **annotation** and **reflection** implementation is useful for user or not, I hide these two method to private.

Since all streamApp only has an entry point : `com.island.ohara.streams.StreamApp`, The configs could be fetch by `docker run .... STREAM_ENTRY "{Some internal used params}"`
Although it looks a little "injection smell", we can enhance this in the future by an embedded REST server if there were lots features needed to be exposed.

saivirtue: > Perhaps we should disable the changeable version of deps.

Does this mean we should fix ALL dependencies version until compatible breaking?

chia7712: > Does this mean we should fix ALL dependencies version until compatible breaking?

yep :)

chia7712: > I am not intend to expose the reflection class to user, and it should not too.
The whole idea is to open a door for user to define custom configuration by method ConfigDef config(). Since I'm not sure the annotation and reflection implementation is useful for user or not, I hide these two method to private.

So ... why we need reflection and annotation if you plan to expose custom configs via ConfigDef config()?

BTW, why not reusing setting definitions in connector? It is powerful than ConfigDef.

saivirtue: **Reflection** could reduce the duplicate code for using internal definition (like default configs).
for example, if we define a config like `val myConfig = ...`, and need to add this to the default configs  set, we will need to add this variable to all needed class.

**Annotation** this is just my favorite to define "metadata" of configs since annotation is born to be used for metadata ;)
But for code consistency, I'm ok to replace this to fields.

`SettingDefinition` is really really powerful, but I think we should have some consensus of **Reflection** before reuse this class (since there are some connector dependency logic in this class)

chia7712: > for example, if we define a config like val myConfig = ..., and need to add this to the default configs set, we will need to add this variable to all needed class.

There are many way to "set" default value to any class. Why you prefer to use reflection. I'd like reflection for its unparalleled flexibility of calling any bit of class. However, it produces many side effect which can hurt us in developing and complicate our code base. In short, it would be better to have a irresistible reason to use reflection :)

saivirtue: Currently, there is only one reason to use reflection : annotation.
I think this is not the "irresistible" reason :)
I will push another commit to refactor `ConfigDef` to `SettingDefinition`

saivirtue: The next commit will included :
- [ ] move refactor of `SettingDefinition` to another issue (#1765)
- [ ] Change the dependencies version be fixable
- [ ] use the common `SettingDef` for streamApp

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
```diff
@@ -0,0 +1,325 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.island.ohara.common.annotations.Nullable;
+import com.island.ohara.common.annotations.Optional;
+import com.island.ohara.common.json.JsonObject;
+import com.island.ohara.common.json.JsonUtils;
+import com.island.ohara.common.util.CommonUtils;
+import java.util.*;
+
+/** This class is used to define the setting of ohara object. */
+public class SettingDef implements JsonObject {
```
chia7712: Do we have a issue of refactoring this one?

saivirtue: Sorry, I had commented before and mark resolved :(
please see #1765

saivirtue:
retry configurator

saivirtue:
> Please file another issue to introduce the custom configs of streamapp

Sure :)

saivirtue:
retry it

saivirtue:
retry configurator

konekoya:
retry manager-api

saivirtue:
retry

saivirtue:
the failed test is not related, merge now

Issue #1809, opened 2019-07-10
Refine Node APIs


chia7712:
The Node APIs is the entrypoint to marshal our distributed envs. With more collie impls and apps, the node interface gets stale and weird. This issue aims to refine Node APIs and it consists of following issue.

- [x] remove --node from Configurator (#1835)
- [x] add "hostname" to node APIs (#1817)
- [x] k8s mode should be able to ignore user name and password (#1859)

chia7712:
All tasks are completed!

PR #1817, opened 2019-07-10
#1809 add "hostname" to node APIs

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1809)


docs/src/rest_interface.md
```diff
@@ -1309,7 +1309,8 @@ The version (tag) depends on which ohara you used. It would be better to use the
 the version of ohara configurator you are running is 0.4, then the official images you should download is oharastream/xx:0.4.
 
 The properties used in describing a node are shown below.
-1. name (**string**) — hostname of node
+1. name (**string**) — name of node
+1. hostname (**string**) — hostname of node
```
saivirtue: So, the "actual" node name is the field **hostname** ?

chia7712: > the "actual" node name is the field hostname ?

yep. the "name" to which collie connects is "hostname" than "name". This change decouples the "hostname" from "name" so the rules to name in the future won't hurt "hostname".

saivirtue: > the "name" to which collie connects is "hostname" than "name". This change decouples the "hostname" from "name" so the rules to name in the future won't hurt "hostname".

Maybe we need some description to avoid user confuse these two fields :)

chia7712: got it

chia7712:
I don't want to add a extra "hostname" to node APIs since the hostname could confuse us in connecting to remote node. However, I do like to add hostname to be alias in Restful APIs and scala APIs as "hostname" is more straightforward to me.

chia7712:
retry it

chia7712:
unrelated error

Issue #1915, opened 2019-07-18
Should separate collie creation to two stages also


saivirtue:
After #1869, zookeeper creation need to use two steps (create -> start). We need apply this rule into ohara-it also.

PR #1916, opened 2019-07-18
 #1915 Should separate collie creation to two stages also

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1915)


ohara-agent/src/main/scala/com/island/ohara/agent/Collie.scala
```diff
@@ -135,6 +135,13 @@ trait Collie[T <: ClusterInfo, Creator <: ClusterCreator[T]] {
     * @return true if it does remove a node from a running cluster. Otherwise, false
     */
   def removeNode(clusterName: String, nodeName: String)(implicit executionContext: ExecutionContext): Future[Boolean]
+
+  // API will need two stages to start a cluster (create -> start),
+  // but collie does not need to store "cluster information" for creation.
+  // We just add a "return cluster" function for testing only
```
chia7712: we should NOT add code which is used in testing only to interface.

perhaps we can replace the impl of zk_start by "cluster(clusterName).map(_._1)"

saivirtue: Sure.

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry

saivirtue:
retry it

Issue #1912, opened 2019-07-17
Support timestamp in metrics


saivirtue:
Since metrics from jmx service is not sync to get (i.e, it may delay for each data in metrics)
we should support timestamp

PR #1919, opened 2019-07-18
#1912 Support timestamp in metrics

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1912)


chia7712:
What is means "startTime" for metrics? The time of generating the record in remote machine? Or the time to generate our BeanObject ?

saivirtue:
> What is means "startTime" for metrics? The time of generating the record in remote machine? Or the time to generate our BeanObject ?

It should be the time of the record generated in remote machine. Let me add some document.

chia7712:
> It should be the time of the record generated in remote machine.

Do all remote metrics objects have “startTime”? The metrics objects which are not generated by our APIs may have no startTime.

saivirtue:
> Do all remote metrics objects have “startTime”? The metrics objects which are not generated by our APIs may have no startTime.

Do you mean like topic metrics? In that case, we use the BeanObject created time instead.
Since our metrics API allow user to define the "startTime" of each metric, Maybe we should remove the changeable "startTime" field to get the consistency?

chia7712:
> Do you mean like topic metrics? In that case, we use the BeanObject created time instead.

It means the "startTime" is not the time of creating the metrics object, right? It causes the startTime from our metrics rest api is always the time of creating BeanObject. For example,

first call: GET /v0/topics/aaa
```json
{
  "startTime": 123
}
```

second call: GET /v0/topics/aaa
```json
{
  "startTime": 124
}
```

The "startTime" in both call should be identical since "startTime" is the time of "creating" metrics object. Personally, it is a critical bug to our rest apis.

> Since our metrics API allow user to define the "startTime" of each metric, Maybe we should remove the changeable "startTime" field to get the consistency?

The changeable startTime is useful in cloning object.

saivirtue:
The "startTime" in both call should be identical since "startTime" is the time of "creating" metrics object. Personally, it is a critical bug to our rest apis.

Indeed. Let me find out if there were any native time of metric data could be used.

The changeable startTime is useful in cloning object.

OK

Issue #1734, opened 2019-07-04
add group to all objects


chia7712:
from #1546

1. - [x] enable DataStore to save group and name (#1927)
1. - [x] rename "FileKey" to "DataKey" (#1928)
1. - [x] replace RocksDataStore.Key by DataKey (#1948)
1. - [x] add basic group handle to basic route (#1950)
1. - [x] add group to ftp route (#1952) 
1. - [x] add group to hdfs route (#1953) 
1. - [x] add group to jdbc route (#1962) 
1. - [x] add group to pipeline route (#1964) 
1. - [x] add group to stream route (@saivirtue ) (#2071)
1. - [x] add group to response of topic (#1965)
1. - [x] add group to response of connector (#1966)
1. - [x] change the flow of pipeline from "string" to "DataKey" (#2002)
1. - [x] the url generator should be cognizant of group and name (#2003)
1. - [x] make ClusterInfo be sub class of Data (#2006)
1. - [x] enable group for Topic (#2011)
1. - [x] add metadata "TOPIC_KEYS" type (#2016)
1. - [x] enable group for Connector (#2029)
1. - [x] enable default group in parsing "ObjectKey" and "TopicKey" (#2056)
1. - [x] move group from query parameter to payload of Creation (#2057)
1. - [x] move "default group" to each APIs (#2061)
1. - [x] Remove name-related methods from DataStore (#2074)
1. - [x] The Client APIs should substitute Key for name (#2077)
1. - [x] add group to Object APIs (#2096)

chia7712:
@konekoya @wu87988622 除了cluster相關的服務外（zk, bk, wk, streamapp and shabondi）其他物件的操作都已經上group了，這邊需要前端配合在查詢的時候搭配group來查詢相關物件的資料。

我們後續的PR需要等到前端採用group後才可以繼續進行。此外以下是group例外的總和:
1. node不支援group
1. topic的group"目前"是default (我們最終會將其改成跟brokerClusterName相同)
1. connector的group"目前"是default (我們最終會將其改成跟workerClusterName相同)
1. zk不支援group
1. bk不支援group
1. wk不支援group
1. streamapp"目前"不支援group

chia7712:
I plan to change the location of group in Create request.

before
```
POST /v0/$rs?group=${g0}
{
  "name": "n0"
}
```

after
```
POST /v0/$rs
{
  "group": "g0",
  "name": "n0"
}
```

[...]

PR #1927, opened 2019-07-18
#1734 enable DataStore to save group and name

chia7712:
### Checklist
- [x] DON'T set a Milestone


ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
```diff
@@ -76,6 +76,8 @@ object StreamApi {
                                         lastModified: Long,
                                         tags: Map[String, JsValue])
       extends Data {
+    // streamapp does not support to define group
```
saivirtue: This group equals to "FileKey group" for current implementation ; We need to bind a jar to this object which is uploaded in workspace.
It is ok to set default value :)

chia7712: > This group equals to ...

Did you mean that the (default value of) group of a streamapp should be equal to group of jar?

saivirtue: > Did you mean that the (default value of) group of a streamapp should be equal to group of jar?

I had to say, for now.
For example,
```
{
  "fileKey": {"name": "xxx.jar", "group": "abc"}
  "group": "bar"
}
```
it is useless that we assign group to "bar" instead of "abc" since streamApp doesn't need to be "group", it is an application.
So it is ok to assign "group=abc" in current implementation.

chia7712: We are on the same page. I prefer to decouple the group in streamapp and group in jar (file) ... :)

Issue #1765, opened 2019-07-08
Refactor setting definition of connector


saivirtue:
For the following reasons, we need to define a basic setting class `common.setting.SettingDef` and refactor the connector setting in this issue:
1. `Builder` must extend pattern ; we cannot abstract common builder class
2. `SettingDefinition` has a field `checker` which is unused for serializing / deserializing
3. `jackson.fastxml` is hard to used in inherited class constructor

the following tasks need to be done:
- [x] Refactor `common.setting.SettingDef` to support checker (#2063)
- [x] Add checker to `StreamDefinitions` (move to #2198)

PR #1929, opened 2019-07-19
#810 Support custom config for stream API

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #810)


chia7712:
what is the main difference between SettingDefinition, SettingDef and StreamDefinition? There are a bunch of duplicate code between SettingDefinition and SettingDef. Could we re-arch them before applying?

saivirtue:
> what is the main difference between SettingDefinition, SettingDef and StreamDefinition? There are a bunch of duplicate code between SettingDefinition and SettingDef. Could we re-arch them before applying?

SettingDefinition is used for connector as it born to, and it has a little difficult to refactor (The work is move to #1765)
SettingDef is used to replace SettingDefinition after #1765, let me address some comment.

In the other way, StreamDefinition is a "name" of streamApp definition (like `ConnectorDefinition`), which combine (className, List<SettingDef>)

I think we need another naming for `StreamDefinition`

chia7712:
> SettingDefinition is used for connector as it born to, and it has a little difficult to refactor (The work is move to #1765)
SettingDef is used to replace SettingDefinition after #1765, let me address some comment.

ok

> In the other way, StreamDefinition is a "object" of streamApp definition (like ConnectorDefinition), which contains (className, List<SettingDef>)I think we need another naming for StreamDefinition

In addition to code, we should avoid to have "similar" naming  for "similar" class. For instance, could we have a class consisting of "className" and "settings", and then make it be used by both connector and streamapp? ConnectorDefinition and StreamDefinition are on the API layer. Unifying them can eliminate duplicate stuff, and another merit of unifying is that it "friendly" enforce us to keep both streamapp and connector have consistent interface to restful api.

saivirtue:
> In addition to code, we should avoid to have "similar" naming for "similar" class.

That seems fair and more clear, let me do it right away :)

konekoya:
@saivirtue I'll give this a look this afternoon. Let me know when the merge conflict is resolved :)

saivirtue:
@konekoya feel free to take over this :)

konekoya:
@saivirtue Please resolve the conflict :-)

saivirtue:
retry connector

saivirtue:
retry manager-api

docs/src/rest_interface.rst
```diff
@@ -3398,27 +3398,48 @@ re-download everything when running services.
...
-        - meters[i].value (**double**) — the number stored in meter
-        - meters[i].unit (**string**) — unit for value
-        - meters[i].document (**string**) — document of this meter
-        - meters[i].queryTime (**long**) — the time of query metrics from remote machine
-        - meters[i].startTime (**option(long)**) — the time of record generated in remote machine
-#. exactlyOnce (**boolean**) — enable exactly once
-#. error (**option(string)**) — the error message from a failed streamApp.
-   If the streamApp is fine or un-started, you won’t get this field.
-#. lastModified (**long**) — last modified this jar time
+.. _restinterface-streamapp-stored-data:
```
vitojeng: `_restinterface-streamapp-stored-data:`
should be `_rest-streamapp-stored-data:`

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/Definition.scala
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+  * The definition of a connector or streamApp.
+  * It should consist of a class name and SettingDef list.
+  *
+  * @param className class name of this object
+  * @param settingDefs definitions of this object
+  */
+//TODO Since we have two type of definition in SettingDef and SettingDefinition
+//TODO remove this after #1765...by sam
```
chia7712: Should we refactor setting definition before adding new feature to streamapp? I just worry there are too many "intermediate idea" in this PR, and it is too cryptic to review code :(

saivirtue: That is really a hard work to do but must do... I will try to work on #1765 before @konekoya finish the frontend part :(

chia7712: another way is to merge it if no obvious error. And complete the refactor ASAP :)

saivirtue: > another way is to merge it if no obvious error.

The obvious error is "broken manager-api"...

chia7712: @konekoya @wu87988622 there is a party to you again :)

konekoya: Yep, we are working on some bugs in the pipeline, and will start to work on this PR once we're done with the pipeline work.

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
```diff
@@ -126,67 +138,147 @@ public String get(String name) {
...
+            .key("servers")
             .group(CORE_GROUP)
             .displayName("Broker List")
             .readonly()
             .documentation("The broker list of current workspace")
+            .valueType(Type.ARRAY)
             .build();
 
-    private static SettingDef APPID_DEFINITION =
+    public static SettingDef IMAGE_NAME_DEFINITION =
```
chia7712: final

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
```diff
@@ -126,67 +138,147 @@ public String get(String name) {
    *
    * @return config object list
    */
-  @JsonGetter
-  List<SettingDef> values() {
+  @JsonGetter(ROOT)
```
chia7712: ROOT?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
```diff
@@ -48,99 +50,176 @@ object StreamApi {
...
     // streamapp does not support to define group
     override def group: String = GROUP_DEFAULT
+    override def name: String = plain(DefaultConfigs.NAME_DEFINITION.key())
     override def kind: String = STREAM_SERVICE_NAME
     override def ports: Set[Int] = Set(jmxPort)
+    override def tags: Map[String, JsValue] = noJsNull(settings)(DefaultConfigs.TAGS_DEFINITION.key()).asJsObject.fields
+
+    def imageName: String = plain(DefaultConfigs.IMAGE_NAME_DEFINITION.key())
+    def instances: Int = plain(DefaultConfigs.INSTANCES_DEFINITION.key()).toInt
+    def jarKey: Option[ObjectKey] =
```
chia7712: Is it illegal to have a StreamClusterInfo without jar key?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/Definition.scala
```diff
@@ -0,0 +1,58 @@
...
+  * @param definitions definitions of this object
+  */
+private[ohara] final case class Definition(className: String, definitions: Seq[SettingDef])
+
+object Definition {
+  private[this] val CLASS_KEY: String = "className"
+  private[this] val SETTING_DEF_KEY: String = "definitions"
+
+  implicit val DEFINITION_JSON_FORMAT: RootJsonFormat[Definition] = JsonRefiner[Definition]
+    .format(new RootJsonFormat[Definition] {
```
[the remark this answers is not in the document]
saivirtue: You are right:)

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
saivirtue: the json object "root" element field name.
Let me change the naming and make some comment.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
saivirtue: Yes. But the jarKey could be changed during UPDATE and it can be None during CREATE. So I move the checking to action phase.

chia7712: The cluster info is the final "info" and it should be always "completed". Otherwise, it causes a weird scenario that a cluster info returned from collie and it "may" lack of jar key?

saivirtue: > a cluster info returned from collie and it "may" lack of jar key?

This should not happened. collie will return the info by actual container, and if a container was successful running, the jarKey exists.

I choose an option object here is because user may create an info with lack jarkey, we cannot auto-filled this field for user.

chia7712: My point is that the field is declared as "option". It means the cluster info is optional. That is what interface means. For another reason, how to take the defined value actually? jarKeys.get? How to "tell" people that the cluster info "must" carry the jarKey?

Rejecting the creation lacking of jarKey won't hurt us. Or changing the interface from option[xxx] to xxx is ok as well (you can produce a error with complete explanation when there is no value).

saivirtue: OK.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
```diff
@@ -48,99 +50,178 @@ object StreamApi {
...
+      .map {
+        case s: JsObject => s.fields
+        case other: JsValue =>
+          throw new IllegalArgumentException(s"the type of tags should be JsObject, actual type is ${other.getClass}")
+      }
+      .getOrElse(Map.empty)
+
+    override def ports: Set[Int] = Set(plain(DefaultConfigs.JMX_PORT_DEFINITION.key()).toInt)
+
+    def jarKey: Option[ObjectKey] =
```
chia7712: 其實可以練習一下如何做向後相容的修改，有點麻煩和痛苦，但是之後必備的技能

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
```diff
@@ -48,99 +50,178 @@ object StreamApi {
...
+
+    def instances: Int = plain(DefaultConfigs.INSTANCES_DEFINITION.key()).toInt
   }
   implicit val STREAM_CREATION_JSON_FORMAT: OharaJsonFormat[Creation] =
     JsonRefiner[Creation]
-      .format(jsonFormat9(Creation))
+      .format(new RootJsonFormat[Creation] {
+        override def write(obj: Creation): JsValue = JsObject(noJsNull(obj.settings))
+        override def read(json: JsValue): Creation = Creation(json.asJsObject.fields ++
+          // using extra field to overwrite the pre-defined fields
```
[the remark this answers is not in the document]
saivirtue: please see [docs](https://github.com/saivirtue/ohara/blob/fix_810/docs/src/rest/rest-stream.rst#create-properties-of-specific-streamapp)

Since "settings" is used for user to define custom key-value, we cannot restrict user to "can or cannot" do. So I prefer to let user decide and remind that "if something is defined before, the "field" in key "settings" will overwrite it.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
saivirtue: 這邊的意思是?

saivirtue:
retry it

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
chia7712: 就是試著去兼容舊的APIs之下加入新的APIs

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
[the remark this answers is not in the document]
chia7712: so there is a "field" called "settings"? The custom stuff to streamapp are under "settings"?
for example:
```json
{
  "name":"a",
  "settings": {
    "k0": "v0"
  }
}
```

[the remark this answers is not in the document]
saivirtue: Yep.

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
saivirtue: OK. 但這次與 @konekoya 一起做這隻PR是要改成使用metadata，改動幅度大，就沒有做向後相容了
之後如果有API change會特別注意這件事

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/StreamApi.scala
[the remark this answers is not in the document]
chia7712: It introduces a bit issue - Ohara manager can't totally depend on SettingDef to generate the form. Ohara Manager need to hardcode the "name", "imageName", and so on. By contrast, the other objects exercising SettingDef put all fields on SettingDef and hence Manager can form the page dynamically and totally.

FYI @konekoya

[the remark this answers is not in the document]
saivirtue: > Ohara manager can't totally depend on SettingDef to generate the form. Ohara Manager need to hardcode the "name", "imageName", and so on

不太了解你的意思。以你的例子來說，比較完整的例子是
```
{
  "name":"a",
  "node": "n1",
  "settings": {
    "name": "abc",
    "k0": "v0"
  },
  "definitions": [
    {"name": ....}
    {"imageName": ...}
  ]
}
```

這裡的 `settings.name` 會覆蓋上一層的 `name` (API最後會得到"name": "abc")
name": "abc")
而manager只需要看`definitions`，SettingDef事先定義好的的"key"都會在"settings"內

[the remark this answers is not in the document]
chia7712: what we are discussing is "creation", right?

It would be better the available fields in "creation" is what the setting definitions declares.

For example, the setting definitions of "xxx streamapp" is 
```json
[
  {
    "key":"name",
    "type":"string"
  },
  {
    "key":"port",
    "type":"int"
  }
]
```

the supposed fields for "creation" is 
```json
{
  "name": "value0",
  "port": 22
}
```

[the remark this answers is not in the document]
saivirtue: > what we are discussing is "creation", right?

Yes. But I think what I point out is "response"

> For example, the setting definitions of "xxx streamapp" is

The description you mention here is right except the fields for "creation" is
```
{
  "settings": {
    "name": "value0",
    "port": 22
  }
}
```
The purpose of adding a "settings" field is that I want to separate the "definition value" and "cluster description" (You can see the structure in `StreamClusterInfo`).
And for the previous overwrite problem, since I cannot know the "custom setting" (which will be extract from custom jar) fields ; and I have to decide if some of the custom setting is used for "cluster description", whether overwrite or not.

creation example
```
{
  "settings": {
    "name": "aa",
    "nodeNames": "user_want_value"
  },
  "nodeNames": ["n1"],
  "metrics": []
}
```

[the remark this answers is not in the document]
chia7712: > creation example

IIRC, it does not have "metrics".

> The purpose of adding a "settings" field is that I want to separate the "definition value" and "cluster description"

How ohara manager to know which settings are "cluster description"?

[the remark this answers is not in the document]
saivirtue: > IRC, it does not have "metrics".

Just example, please ignore that.

> How ohara manager to know which settings are "cluster description"?

the process of manager:
get definitions => render metadata form => POST a creation request :
1. the value of metadata is belong to "settings", i.e., definition value
2. the value not present on manager is belong to cluster description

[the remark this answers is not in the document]
chia7712: > the value not present on manager is belong to cluster description

Does Manager need to hardcode for the fields of this part?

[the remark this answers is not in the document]
saivirtue: > Does Manager need to hardcode for the fields of this part?

No needed :)
Since this part is from collie, our collie is strong enough to handle it.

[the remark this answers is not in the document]
chia7712: Does manager understand the “keys” of “cluster description” (for example, name, imageName, nodeNames and so on) if those “keys” are not included to SettingDef?

Following above issue. Could we add an new field to “cluster description” and the new field is able to display on UI even if we do not do any change to the UI code?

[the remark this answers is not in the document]
saivirtue: > Does manager understand the “keys” of “cluster description” (for example, name, imageName, nodeNames and so on) if those “keys” are not included to SettingDef?

The keys of cluster description are only shown in cluster info, which is the response.
In creation, user (or manager) doesn't need to care the keys of cluster description, they only need to know is the keys of "definitions"

> Following above issue. Could we add an new field to “cluster description” and the new field is able to display on UI even if we do not do any change to the UI code?

I think here you mean "name", "imageName", "nodeNames", right?
name and imageName are part of SettingDef in stream, so they are under control. (display on UI)

nodeNames is not the field that manager could change, manager use `instances` instead (hence settingDef doesn't contain nodeNames definition), but if user uses API and really want to change the nodeNames:
```
{
  "settings": {
    "nodeNames": ["n1", "n2"]
  }
}
```
after UPDATE, the nodeNames we saved will be overwrite.

use "cluster description" and "definition value" to describe the situation is a little weird, I'd to say.
I just want to describe those not defined in SettingDef fields (for example, nodeNames) are not shown on UI because those fields are used internally.
After the inspection, these fields are:
1. nodeNames
2. deadNodes
3. state
4. error
5. metrics
6. lastModified
7. group (stream doesn't support group now)
8. kind
9. ports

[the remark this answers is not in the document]
saivirtue: use "cluster description" and "definition value" to describe the situation is a little weird, I'd to say.
I just want to describe those not defined in SettingDef fields (for example, nodeNames) are not shown on UI because those fields are used internally.
After the inspection, these fields are:
1. nodeNames
2. deadNodes
3. state
4. error
5. metrics
6. lastModified
7. group (stream doesn't support group now)
8. kind
9. ports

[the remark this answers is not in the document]
saivirtue: some follow-up tasks are defined on #2191.
I will work on the following changes for this PR:
1. remove the "settings" key in creation
2. move non-generated fields from outside to "settings" in clusterInfo

Issue #1905, opened 2019-07-17
Zookeeper start should be idempotent


saivirtue:
start zookeeper should return same result (no matter how many times we send the request)

Issue #1544, opened 2019-06-21
Separate cluster creation into two phase


saivirtue:
We need to handle the cluster creation in a stage way :
1. store the request properties in configurator
2. create cluster

As #1942 discussed, the following issue should satisfy:
<details open><summary>requirement list</summary>

- create zookeeper data: POST with payload, response created object
- start zookeeper: PUT without payload, response httpcode=202
- stop zookeeper: PUT without payload, response httpcode=202
- delete zookeeper data: PUT without payload, response httpcode=204
- get zookeeper information: GET without payload, response up-to-date object (include status)

</details>

Since this issue change ALL cluster APIs, we separate to following tasks:
- [x] Separate cluster creation of zookeeper (#1869)
-- Zookeeper start should be idempotent (issue #1905)
- [x] Separate cluster creation of broker (#1971)
- [ ] Separate cluster creation of worker (#2144)
-- move auto-filled logic to creation of worker route (added from #2005)
- [x] ~Separate cluster creation of stream~ stream cluster is already two-phase creation
-- Integrate StreamAppDescription into StreamClusterInfo (issue #1025

chia7712:
@saivirtue Auto-completing properties in START phase produces a embarrasses scenario that users can’t observer what is used to build their services. Hence, users will get a cluster they don’t like.

User is able to see the changes of properties if we auto-complete properties in CREATE phase, because the response is the properties we have updated.

Any feedbacks?

saivirtue:
> Auto-completing properties in START phase produces a embarrasses scenario that users can’t observer what is used to build their services. Hence, users will get a cluster they don’t like.

We already add some description about auto-completed parameters in docs.
I think this is a trade-off situation: if you don't like so many parameters, you should care "the default" value.

chia7712:
> if you don't like so many parameters, you should care "the default" value.

user CANNOT care the default value before starting if we auto-complete the properties in start phase And I don't think the docs is a "good" place to read the latest default value :(

saivirtue:
> user CANNOT care the default value before starting if we auto-complete the properties in start phase 

Sure. User need to request another GET to get the value.
But as I worried about, if auto-completed was happened in creation phase:
step 1. absent parameter in creation -> response auto-completing
step 2. start this object, and then stop
step 3. update auto-completed parameter -> response should override the parameter?

If no, that is a weird response, so skip it.
if yes, we will override the data in configurator, and start action again will run in a different situation.

for example, 
```
[...]

PR #1935, opened 2019-07-19
#1905 Zookeeper start should be idempotent

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1905)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/RouteUtils.scala
```diff
@@ -161,11 +161,18 @@ private[route] object RouteUtils {
...
-              .value[Req](clusterName)
-              .flatMap(req =>
-                if (req.nodeNames.isEmpty) throw new IllegalArgumentException(s"You are too poor to buy any server?")
-                else basicCheckOfCluster2(nodeCollie, clusterCollie, req).map(clusters => hookOfStart(clusters, req))))
+            complete(store.value[Req](clusterName).flatMap { req =>
+              collie.exist(req.name).flatMap {
+                if (_) {
+                  Future.successful(req)
+                } else {
+                  if (req.nodeNames.isEmpty)
```
chia7712: Should we move this check to parse phase?

saivirtue: For any cluster, a `start` command should be idempotent. This will enable cluster start to be idempotent if it extended this method.
Do you mean this?

chia7712: Pardon me :( 

My point is that route should not check the **empty**, **negative** or **null** as much as possible ...

wu87988622:
@saivirtue 
In the fake mode, I first start the zookeeper and then get the zookeeper, no matter how long wait, I won't get any "RUNNING" status, but in k8s mode there is no such problem,

chia7712:
> In the fake mode, I first start the zookeeper and then get the zookeeper, no matter how long wait, I won't get any "RUNNING" status, but in k8s mode there is no such problem,

I think @saivirtue needs some time to address the result we had discussed.

wu87988622:
> > In the fake mode, I first start the zookeeper and then get the zookeeper, no matter how long wait, I won't get any "RUNNING" status, but in k8s mode there is no such problem,
> 
> I think @saivirtue needs some time to address the result we had discussed.

OK Thank you~

konekoya:
@saivirtue Since @wu87988622 is now busy with #1961 I'll be taking care of this issue :)

ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestZookeeperRoute.scala
```diff
@@ -223,6 +224,29 @@ class TestZookeeperRoute extends MediumTest with Matchers {
...
   }
 
+  @Test
+  def testIdempotentStart(): Unit = {
+    val zk = result(
+      zookeeperApi.request.nodeNames(nodeNames).create()
+    )
+    result(zookeeperApi.start(zk.name))
+    val info1 = result(zookeeperApi.get(zk.name))
+    // duplicated start will return the current cluster info
```
konekoya: Is this still the case? Or it will return the HTTP status code again?

saivirtue: As #1544 state:
```
start zookeeper: PUT without payload, response httpcode=202
get zookeeper: GET without payload, response up-to-date object (include status)
```

The `start` api will have no response (only httpCode=202)
And the `get` api will return the latest information

saivirtue:
Thanks @konekoya @wu87988622 :)

Issue #1943, opened 2019-07-19
Topic timestamp of metrics should be different from BeanObject.startTime


saivirtue:
Since the kafka metrics does not have timestamp, we should change the "meaning" to another field: `queryTime`. For now, metrics API will have the following fields (the `startTime` is optional):
```json
{
  "value": 1234,
  "unit": "rows",
  "document": "number of processed rows",
  "queryTime": 1563429505055,
  "startTime": 1563429590505
}
```

PR #1959, opened 2019-07-22
#1943 Topic timestamp of metrics should be different from BeanObject.…

saivirtue:
…startTime

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1943)


ohara-metrics/src/main/java/com/island/ohara/metrics/BeanObject.java
```diff
@@ -111,16 +110,15 @@ public Builder attributes(Map<String, Object> attributes) {
       return this;
     }
 
-    @Optional("default is current time")
-    public Builder startTime(long startTime) {
-      this.startTime = startTime;
+    public Builder queryTime(long queryTime) {
+      this.queryTime = queryTime;
```
chia7712: Could we forbid user to input negative number?

saivirtue: Already done :)
Let me add some tests, thanks!

PR #1971, opened 2019-07-23
#1544 Separate cluster creation of broker

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1544)


konekoya:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
```diff
@@ -171,12 +171,17 @@ trait BrokerCollie extends Collie[BrokerClusterInfo, BrokerCollie.ClusterCreator
                     val clusterInfo = BrokerClusterInfo(
                       name = clusterName,
                       imageName = imageName,
-                      zookeeperClusterName = zookeeperClusterName,
+                      zookeeperClusterName = Some(zookeeperClusterName),
```
chia7712: there is an critical issue we have to reach the consensus - which phase we should auto-adjust the property for cluster? For example, should we auto-add the zookeeper cluster to properties if user ignore it in storing properties? Or the phase of starting a broker cluster is more suitable? 

Currently, ConnectorRoute takes the former. Seems the later is more acceptable to us now since the properties stored in configurator is able to be "non-complete". The auto-adjusting produces a error if the properties lacks of cluster name, and hence making auto-adjusting happen on store phase depart from our guarantee.

Please file a issue to ameliorate it for ConnectorRoute if the later is what you anticipate.

saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
saivirtue: > since the properties stored in configurator is able to be "non-complete".

This is what I expected. Auto-adjusting properties for user in storing-phase has another question: Should we update the auto-adjusting field from update API or not?

At any rate, I'm in same page with the last conclusion :)
> if the later is what you anticipate.

Let me fire another issue to unify route of this behavior.

chia7712: BTW, TopicRoute has similar issue as well.

chia7712: Let me take over topic issue :) see #1972

docs/src/rest_interface.md
```diff
@@ -1740,6 +1740,10 @@ The properties which can be set by user are shown below.
 1. zookeeperClusterName (**String**) — name of zookeeper cluster used to store metadata of broker cluster
 1. nodeNames (**array(string)**) — the nodes running the broker process
 1. deadNodes (**array(string)**) — the nodes that have failed containers of broker
+1. tags (**object**) — the user defined parameters
+1. state (**option(string)**) — only started/failed broker has state (RUNNING or DEAD)
+1. error (**option(string)**) — the error message from a failed broker. If broker is fine or un-started, you won't get this field.
+1. lastModified (**long**) — last modified this jar time
```
vitojeng: Just a reminder, transform **rest_interface.md** would be finished today, you may need to update **rest_interface.rst**

saivirtue: got it

konekoya:
@saivirtue Let me know when you're done addressing the above comments. I'll be making the changes that are necessary for the frontend and push again.

ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
saivirtue: Sure :)
The ConnectorRoute rest enhancement is #1978

saivirtue:
retry it

konekoya:
retry manager-it

konekoya:
@saivirtue Could you help resolve the conflict? After that, I will retry manager's it again :)

saivirtue:
/cc @konekoya 
It's ok now :)

saivirtue:
retry it

saivirtue:
retry it

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ClusterAccess.scala
```diff
@@ -17,27 +17,34 @@
...
     extends BasicAccess(prefixPath) {
+  val START_COMMAND: String = "start"
+  val STOP_COMMAND: String = "stop"
 
   private[this] def _clusterName(name: String): String =
     CommonUtils.requireNonEmpty(name, () => "cluster name can't be empty")
-  private[this] def _nodeName(name: String): String =
+  private[this] def _nodeName(name: String): String = {
+    //TODO: This is a workaround, but we should apply jsonRefiner to here...by Sam
+    if (name == START_COMMAND || name == STOP_COMMAND)
```
chia7712: This method is located at client-side, right? Rejecting the illegal word is the prerogative of server-side, and client-side needs not do it.

ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
```diff
@@ -324,7 +341,9 @@ object BrokerCollie {
     private[this] var jmxPort: Int = CommonUtils.availablePort()
 
     override protected def doCopy(clusterInfo: BrokerClusterInfo): Unit = {
-      zookeeperClusterName(clusterInfo.zookeeperClusterName)
+      zookeeperClusterName(
```
chia7712: How about coping it only if the input bk cluster has zk cluster?

saivirtue: You mean `doCopyOrDoNothing()` ?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ClusterAccess.scala
saivirtue: OK

ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
chia7712: yep

saivirtue: I think we can "fail fast" in here ; Since `doNothing` will cause exception in actual action (in here, it's addNode). 
Since this method is actual "do copy from actual cluster", Maybe absent zk cluster case not exists?

chia7712: You are right :)

konekoya:
retry manager-it

saivirtue:
retry it

konekoya:
retry manager-it

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry it

konekoya:
retry manager-it

Issue #1978, opened 2019-07-24
Move data-checking of connectorRoute to action phase


saivirtue:
As [link](https://github.com/oharastream/ohara/pull/1971#discussion_r306234487) described, data stored in configurator could be checked in action phase.
Include:
- move checking `workerClusterName` from create phase to start phase
- update data of configurator in `update()` method

PR #1991, opened 2019-07-25
#1978 Move data-checking of connectorRoute to action phase

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1978)


saivirtue:
> Have we added the workerClusterName to connector description if workerClusterName is ignored by user?

In create/update phase, no.
In start phase, we will check the absent cluster name, and update the data of configurator.

ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestConnectorRoute.scala
```diff
@@ -101,19 +112,29 @@ class TestConnectorRoute extends SmallTest with Matchers {
...
+    val topic = result(
+      TopicApi.access
+        .hostname(configurator.hostname)
+        .port(configurator.port)
+        .request
+        .name(CommonUtils.randomString(10))
+        .create())
+
+    // absent worker cluster is ok since there is only one worker cluster
+    result(connectorApi.request.topicName(topic.name).create().map(info => connectorApi.start(info.name)))
```
chia7712: result(connectorApi.request.topicName(topic.name).create()."flatMap"(info => connectorApi.start(info.name)))

you should flatmap the future chain. otherwise, the last thread is leak from you wait process.

chia7712:
> In start phase, we will check the absent cluster name, and update the data of configurator.

the worker cluster is filled by configurator. However, the workerClusterName is not attached to ConnectorDescription in "starting" phase. Could you have a test for that case?

ohara-configurator/src/test/scala/com/island/ohara/configurator/route/TestConnectorRoute.scala
saivirtue: My bad :(

saivirtue:
> > In start phase, we will check the absent cluster name, and update the data of configurator.
> 
> the worker cluster is filled by configurator. However, the workerClusterName is not attached to ConnectorDescription in "starting" phase. Could you have a test for that case?

The test case is in the [comment](https://github.com/oharastream/ohara/pull/1991#discussion_r307122030) :)

chia7712:
> The test case is in the comment :)

Pardon me. I fail to catch the point.

There is a test of starting a connector without specific worker. You should check the response whether it has worker cluster name. And I believe it does not. For another case, the connector desc stored in configurator does not have worker cluster as well.

In short, the worker name Configurator automatically pick up for connector should be written to connector desc. Otherwise, a starting connector without worker cluster is produced, and it is utter a bug l’d say.

saivirtue:
> In short, the worker name Configurator automatically pick up for connector should be written to connector desc. Otherwise, a starting connector without worker cluster is produced, and it is utter a bug l’d say.

Thanks for pointing out, I've updated this pr and add some check testing.

Issue #2001, opened 2019-07-25
Integrate StreamAppDescriptioninto StreamClusterInfo  


chia7712:
They are too similar to use accurately ...

saivirtue:
This should be refactor after #1544 - Separate cluster creation of stream done.
Let me add some comment, and feel free to take this :)

saivirtue:
I've been doing this issue :)

Issue #1749, opened 2019-07-05
Refactor manager's API tests


konekoya:
The API tests are now all lived in a single giant file making debugging and tweaking tests extremely hard. We should consider separating the tests into smaller files and removing the shared variables they're using now (hence, less coupling).  

- [x] Add broker delete API test and refactor worker API (#2012)
- [x] Broker API (#1971)
- [x] Refactor topic, container and log API tests (#2014)
- [x] Refactor connector API tests (#2028)
- [x] Refactor node and zookeeper API tests (#2034)
- [x] Enable pipeline API tests (#2052)
- [x] Cleanup services after each API test (#2058) 
- [x] Enable validate and add info API tests (#2059)
- [x] Refactor stream and jar API tests (#2110)

konekoya:
@wu87988622 I'm taking over this issue since I've done a little bit refactoring in #1971 already :) and you can continue your work on #1730

PR #2008, opened 2019-07-26
 #2001 Integrate StreamAppDescription to StreamClusterInfo

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2001)


docs/src/rest_interface.rst
```diff
@@ -3707,6 +3708,14 @@ start a StreamApp
 
 **Example Response**
 
+  ::
+
+    202 Accepted
+
+  .. note::
+
+     You should use `Get stream cluster information <#get-information-from-a-specific-streamapp-cluster>`__ to fetch up-to-date status
```
vitojeng: It would be better to use [`:ref:`](https://www.sphinx-doc.org/en/1.5/markup/inline.html#role-ref) instead hard-code link.

saivirtue: OK

saivirtue:
retry common

konekoya:
retry manager-api

konekoya:
Can confirm the test error is unrelated to this PR. The current manager API tests are not very stable. I've filed another PR to fix it in #1749 and I've also tested this PR with the change in #1749 which addresses the unstable issue and the test passed. So I think it's pretty safe to merge this PR :)

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
the failed test case `testWorker` is cause by timeout, unrelated to this PR. Merge it now.

PR #2016, opened 2019-07-28
#1734 add metadata "TOPIC_KEYS" type

chia7712:
### Checklist
- [x] DON'T set a Milestone


ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/DataKey.scala
```diff
@@ -30,10 +31,42 @@ import spray.json.{DeserializationException, JsObject, JsString, JsValue, RootJs
...
+    */
+  def asJava: ObjectKey = ObjectKey.of(group, name)
+
+  /**
+    * generate the topic name for kafka. Noted: kafka topic does not support group so we generate the
+    * name composed of group and name
+    *
+    * @return topic name for kafka
+    */
+  def topicNameOnKafka: String = asJava.topicNameOnKafka()
```
saivirtue: DataKey is a base class for all objects in ohara, and this method seems a little specific. Could we move it to, a sub-class for example, `TopicDataKey` ?

chia7712: you are right :)

Issue #2049, opened 2019-07-30
Redesign connector API's tags definition


konekoya:
Tags are now part of the connector API definition. But needs a bit tweak so it can be properly rendered/ handled in the frontend

- [x] Supply a connector which contains all the valueType that's available to the connector author (@chia7712 ) (#2060)
- [x] Add a new valueType `TAGS` (@chia7712 ) (#2053)
- [x] Render valueType `TAGS` as an text field (#2225)
- [x] Add a tags definition in dumb source and sink connectors (#2224)

PR #2053, opened 2019-07-30
#2049 Add a new valueType "TAGS"

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2049)


chia7712:
> this "tags" is different from Data.tags, right?

Not really. this "tags" is identical to Data.tags. The story here is we try to show the Data.tags via metadata APIs :)

> connector.tags default is null, but Data.tags default is Map.empty
> connector.tags value is string, but Data.tags value is JsValue

the default value of ConnectorDescription#tags is Map.empty[String, JsValue] than null/String.

PR #2063, opened 2019-07-31
 #1765 Refactor common.setting.SettingDef to support checker

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #1765)


saivirtue:
retry

saivirtue:
retry it

saivirtue:
retry it

ohara-kafka/src/main/java/com/island/ohara/kafka/connector/json/ConnectorDefinitions.java
```diff
@@ -0,0 +1,415 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.connect.runtime.rest.entities.ConfigKeyInfo;
+
+/**
+ * This class is used to define the configuration of ohara connector. this class is related to
+ * org.apache.kafka.connect.runtime.rest.entities.ConfigKeyInfo
+ */
+public abstract class ConnectorDefinitions {
```
chia7712: ConnectorDefinition"s", why?

chia7712: Normally, a class having postfix "s" means that it is a collection of helper methods and not instantiatable.

saivirtue: >  it is a collection of helper methods and not instantiatable.

You got my point.
We should initial a "definition" by using SettingDef. This class is more like the old "SettingDefinitions".

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
```diff
@@ -21,13 +21,17 @@
...
+import com.island.ohara.common.exception.OharaConfigException;
 import com.island.ohara.common.json.JsonObject;
 import com.island.ohara.common.json.JsonUtils;
 import com.island.ohara.common.util.CommonUtils;
+import java.io.Serializable;
 import java.util.*;
 
 /** This class is used to define the setting of ohara object. */
-public class SettingDef implements JsonObject {
+public final class SettingDef implements JsonObject, Serializable {
```
chia7712: seems SettingDef  is almost same to SettingDefinition? Why not just move SettingDefinition from ohara-kafka to ohara-common?

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
```diff
@@ -61,9 +66,28 @@
...
+    /**
+     * TAGS is a flexible type accepting a json representation. For example: { "k0": "v0", "k1":
+     * "v1", "k2": ["a0", "b0" ] }
+     */
+    TAGS,
+  }
+
+  /** this class is used to pre-check the setting before running connector. */
+  @FunctionalInterface
+  public interface Checker {
```
chia7712: in #2084, I plan to remove custom checker from setting since a serializable checker is dangerous in spades.

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
chia7712: Moving file is a smoothly change for GIT.

ohara-client/src/test/scala/com/island/ohara/client/configurator/v0/TestConnectorApi.scala
```diff
@@ -222,7 +223,7 @@ class TestConnectorApi extends SmallTest with Matchers {
     val creationRequest = ConnectorApi.CONNECTOR_CREATION_FORMAT.read(s"""
                                                                                       | {
```
Wei-1: One question here. Is there a reason why these Strings are so far back?

chia7712: Seems It is a issue on scalafmt. I'm not sure whether scalafmt allows us to adjust this part manually...

@saivirtue Could you give a help to scalafmt to format code manually?

Wei-1: The scalafmt file defines the maxColumn = 120, but a lot of lines here already exceeded that.
Not sure what will happen when it conflicts with itself.

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
saivirtue: It is because I usually add a new class and put the require part in it without break the current implementation (for testing purpose).
But you are right, I will try to use the "moving" way in next commit.

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
saivirtue: Let me take a look of #2084

ohara-client/src/test/scala/com/island/ohara/client/configurator/v0/TestConnectorApi.scala
saivirtue: > Could you give scalafmt a hand to format code manually?

This is the only thing I can do for current version, I think.

> Not sure what will happen when it conflicts with itself.

Maybe, a explosion? 😆

vitojeng: We really need a more comfortable format. 😄

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
saivirtue: It seems because I had added a `SettingDef` class before...
at any rate, I modified the class to be more like a "copy and paste" job from `SettingDefinition`

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
saivirtue: Since I need to separate some specific type checker (`CONNECTOR_KEY`, `TOPIC_KEYS`) to individual class (ConnectorDefinitions), the custom checker logic is move to ConnectorDefinitions.

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
```diff
@@ -121,6 +154,73 @@ private SettingDef(
...
+                                          + " does not exist in row:"
+                                          + String.join(",", row.keySet()));
+                              }));
+
+            } catch (Exception e) {
+              throw new OharaConfigException(
+                  "the value:" + value + " can't be converted to PropGroups type");
+            }
+            // It is ok to convert the value from string to list<column>, thank God!
+          } else throw new OharaConfigException("the configured value must be string type");
```
chia7712: Excepting for ConfigException, the others cause the server error to kafka. It is great to apply our exception to Checker. However, you have to convert OharaConfigException to ConfigException when you generate kafka checker.

saivirtue: OK

saivirtue:
retry it

saivirtue:
retry it

ohara-kafka/src/main/java/com/island/ohara/kafka/connector/json/ConnectorDefinitions.java
```diff
@@ -0,0 +1,433 @@
...
+          .displayName("Connector key")
+          .key("connectorKey")
+          .valueType(Type.CONNECTOR_KEY)
+          .documentation("the key of this connector")
+          .group(CORE_GROUP)
+          .orderInGroup(ORDER_COUNTER.getAndIncrement())
+          .build();
+
+  @VisibleForTesting
+  static final Consumer<Object> chekerOfConnectorKey =
```
chia7712: Why to extract the inner anonymous class to this static class? The CONNECTOR_KEY is placed at SettingDef, right? Is there no default checker for it?

ohara-kafka/src/main/java/com/island/ohara/kafka/connector/json/ConnectorDefinitions.java
```diff
@@ -0,0 +1,433 @@
...
+          .displayName("Topics")
+          .key("topicKeys")
+          .valueType(Type.TOPIC_KEYS)
+          .documentation("the topics used by connector")
+          .reference(Reference.TOPIC)
+          .group(CORE_GROUP)
+          .orderInGroup(ORDER_COUNTER.getAndIncrement())
+          .build();
+
+  static final Consumer<Object> chekerOfTopicKeys =
```
chia7712: ditto

ohara-kafka/src/main/java/com/island/ohara/kafka/connector/json/ConnectorDefinitions.java
saivirtue: > The CONNECTOR_KEY is placed at SettingDef, right?

yes.

>  Is there no default checker for it?

connector_key has it's own "key class" (the checker depends on it), and move it to default checker is a little weird.

chia7712: SettingDef offers the Type.CONNECTOR_KEY but no default checker for the type?

saivirtue: I am also thinking to move the special type (CONNECTOR_KEY, TOPIC_KEYS) to custom definitions, too. But for the following reasons, I put it to the default place:
1. Since checker is customize-able, we could leave some type with absent checker
2. leave the special type in default place could open a door for other definition (ex: StreamDefinitions) could use the same type
3. Currently, ohara-manager use the type to generate the different form, right?

chia7712: > checker is not customize-able, but some type we cannot define the default checker

the CONNECTOR_KEY and TOPIC_KEY are not included, right? We are able to provide the default checker for them

>  leave the special type in default place could open a door for other definition (ex: StreamDefinitions) could use the same type

We should encourage them (connector, streamapp, topic) to use same type AND same checkers now.

> Currently, ohara-manager use the type to generate the different form, right?

yep. But is this related to the default checker?

saivirtue: Let me summaries:
I prefer not to move the special type checker to default place since we need to refactor a lot of stuff (ex: `ConnectorKey` and `KeyImpl`) to common module (unless we move the default SettingDef to ohara-kafka).
If the later is prefer option (moving SettingDef to ohara-kafka), the problem should be resolved.

chia7712: If connector_key and topic_key are “specific” type to SettingDef,  it would be better to remove them from SettingDef and then have a flexible design to enable sub class to extend the “type” and “check”from SettignDef.

It is ok to leave a issue for this discussion and then go for merge.

saivirtue: see #2124

PR #2077, opened 2019-07-31
#1734 The Client APIs should substitute Key for name

chia7712:
### Checklist
- [x] DON'T set a Milestone


ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/HdfsInfoApi.scala
```diff
@@ -63,6 +63,15 @@ object HdfsInfoApi {
...
 
+    /**
+      * set the group and name via key
+      * @param objectKey object key
+      * @return this request
+      */
+    def key(objectKey: ObjectKey): Request = {
+      group(objectKey.group())
+      name(objectKey.name())
+    }
```
saivirtue: it seems you want to use this instead of `.group().name()`, right?
maybe we need to create an issue to apply this rule to all object ; since we live in an "OjbectKey" world

chia7712: the group is still a optional argument. Keeping the name() method open a room to enable caller to ignore group :)

chia7712:
retry build

chia7712:
retry doc

chia7712:
retry configurator

chia7712:
retry client

Issue #2091, opened 2019-08-02
Move all constants from Data to v0.package


chia7712:
those constants belong to v0 level. Also, the Data should be moved to higher package :)

PR #2094, opened 2019-08-02
#2091 Move all constants from Data to v0.package

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2091)


ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/package.scala
```diff
@@ -16,10 +16,25 @@
...
+    * the default group to all objects.
+    * the group is useful to Ohara Manager. However, in simple case, the group is a bit noisy so we offer the default group to all objects when
+    * input group is ignored.
+    *
+    * This field is protected since it is a "default" to all APIs scopes than a "global" default to whole project.
+    * However, each APIs scope may have custom default value of group.
+    */
+  private[v0] val GROUP_DEFAULT: String = "default"
+  val GROUP_KEY: String = "group"
+  val NAME_KEY: String = SettingDefinition.CONNECTOR_NAME_DEFINITION.key()
```
saivirtue: It is still a little weird that we use "connector name" as object name.

chia7712: > It is still a little weird that we use "connector name" as object name.

It is for consistency. We have to deploy the "name" on metadata and the key of "name" should be same.

saivirtue: got the point :)

Issue #2130, opened 2019-08-05
add helper method to RouteUtils to accept custom action


chia7712:
#2090 is about to add a route to process custom action. The current route utils is not able to address it. As a conscientious utils, RouteUtils should have a way to inject custom route to process any kind of action request.

- [x] a bit refactor for RouteUtils (#2155)

chia7712:
@konekoya @wu87988622 I’m going to remove payload from all action requests. Is  Manager ready to meet this change?

konekoya:
> @konekoya @wu87988622 I’m going to remove payload from all action requests. Is Manager ready to meet this change?

So it's basically, start/stop/delete right? If so, I think it's pretty safe now to remove the payload from them

chia7712:
reopen to do a bit cleanup

PR #2131, opened 2019-08-05
#2130 add helper method to RouteUtils to accept custom action

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2130)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/RouteUtils.scala
```diff
@@ -384,16 +390,75 @@ private[route] object RouteUtils {
...
-        val key = ObjectKey.of(group, name)
-        path(START_COMMAND)(complete(StatusCodes.Accepted -> hookOfStart(key).flatMap(res => store.add[Res](res)))) ~
-          path(STOP_COMMAND)(complete(StatusCodes.Accepted -> hookOfStop(key).flatMap(res => store.add[Res](res)))) ~
-          path(PAUSE_COMMAND)(complete(StatusCodes.Accepted -> hookOfPause(key).flatMap(res => store.add[Res](res)))) ~
-          path(RESUME_COMMAND)(complete(StatusCodes.Accepted -> hookOfResume(key).flatMap(res => store.add[Res](res))))
+        val key = ObjectKey.of(hookOfGroup(groupOption), name)
+        actionHooks
+          .map {
+            case (command, hook) =>
+              path(command)(complete(StatusCodes.Accepted -> hook(key).flatMap(res => store.add[Res](res))))
```
saivirtue: We have a special query parameter `?force=true`, and it should be applied to this basic route in #1544 
could this new structure also support it?

chia7712: roger that :)

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/RouteUtils.scala
```diff
@@ -98,39 +105,68 @@ private[route] object RouteUtils {
...
+    * 2) take finalDeletionHook
+    * 3) return NotFound
     */
-  trait HookOfStop[Res] {
-    def apply(key: ObjectKey): Future[Res]
+  trait ActionHookManager {
+    def deletionHooks: Map[String, ActionHook]
+    def putHooks: Map[String, ActionHook]
+    def finalPutHook: Option[ActionHook]
+    def finalDeletionHook: Option[ActionHook]
```
saivirtue: the "final" hook seems to support the "node action" here.
Just a thought, may be we can change the api to `PUT/DELETE: /v0/{resource}/{name}/node/{nodeName}` ?
It makes the PUT/DELETE actionHook more unified.
We can open an issue to discuss this.

chia7712: Adding/removing node is totally same code between zk, bk, wk and streampp since we have unified them via collie. Hence, I don't expose the hook for handling node. It saves the code for each routes that they don't need to worry about the node changes anymore.

Is there a reason that routes need to do something for node changes?

saivirtue: > Is there a reason that routes need to do something for node changes?

I'm no saying expose the node actions, it is just the un-straightforward naming of "finalXXX".
For node actions, there should be an "action word" like other actions (start, stop...)
But as this is api changes, we can make another discussion with other developers :)

chia7712: > But as this is api changes, we can make another discussion with other developers :)
Given that this PR brings a bit infra to cluster route, we can address the naming in this one :)

> For node actions, there should be an "action word" like other actions (start, stop...)

the finalRoute is not linked to any explicit "action". But, we can have a alias method to handle the "node" explicitly :)

saivirtue: thanks!

konekoya:
retry manager-api

Issue #2231, opened 2019-08-12
Get definition of streamApp should skip quote of url also


saivirtue:
#2214 solved the parsing url problem of streamApp.
However, the definition is still have the same problem...

The proper way to pass the `java.net.URL` is:
`new URI(url.getProtocol, url.getUserInfo, url.getHost...).toASCIIString`

[Oracle URI](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html) explains the detail about "encode each part of url"

we need a proper way to pass the encoded-url to docker.
- [x] use URI constructor instead of URL for parsing url to server (#2248)

chia7712:
If the "quote" is something magic in streamapp, it would be better to lock it in the box than exposing it.

saivirtue:
> If the "quote" is something magic in streamapp, it would be better to lock it in the box than exposing it.

It is not required in streamApp. It's just a missing change in #2214 :(

saivirtue:
It seems still not working, re-open it.

PR #2232, opened 2019-08-12
#2231 Get definition of streamApp should skip quote of url also

saivirtue:
### Checklist
- [x] DON'T set a Milestone


ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
```diff
@@ -154,13 +154,14 @@ trait StreamCollie extends Collie[StreamClusterInfo, StreamCollie.ClusterCreator
...
     */
+  //TODO : this workaround should be removed and use a new API instead in #2191...by Sam
   def definitions(jarUrl: URL): Future[Option[Definition]] =
     Future.successful {
       import sys.process._
       val classpath = System.getProperty("java.class.path")
       val command =
         s"""java -cp $classpath ${StreamCollie.MAIN_ENTRY} ${DefaultConfigs.JAR_KEY_DEFINITION
-          .key()}=\"${jarUrl.toString}\" ${StreamCollie.CONFIG_KEY}"""
+          .key()}=${jarUrl.toString} ${StreamCollie.CONFIG_KEY}"""
```
chia7712: Without quote, the file name having space does not work for streamapp...

saivirtue: you are right. Since our file.name has no limitation...
It is a strange behavior that if the escape char is used in "triple quotes", the `java.net.URL` will complain **no protocol**...

Let me dig further and resolve this.

saivirtue:
retry configurator

Issue #2235, opened 2019-08-12
Stream API could not update from or to field to empty


saivirtue:
We need a way to update the **from** or **to** fields to empty if user want to "disconnect" topics in pipeline

PR #2237, opened 2019-08-12
#2235 Stream API could not update from or to field to empty

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2235)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/StreamRoute.scala
```diff
@@ -188,8 +188,10 @@ private[configurator] object StreamRoute {
           settings = req.settings ++
             Map(
               DefaultConfigs.NAME_DEFINITION.key() -> JsString(key.name),
-              DefaultConfigs.FROM_TOPICS_DEFINITION.key() -> JsArray(req.from.map(JsString(_)).toVector),
-              DefaultConfigs.TO_TOPICS_DEFINITION.key() -> JsArray(req.to.map(JsString(_)).toVector),
+              DefaultConfigs.FROM_TOPICS_DEFINITION.key() -> JsArray(
```
chia7712: the update without "from" will change the previous setting to "Seq.empy". It is a bug, right?

saivirtue: Nope. This the case "use UPDATE as POST". If user request without "from", we generate Set.empty to response.

chia7712: got it :)

saivirtue:
retry client

Issue #2267, opened 2019-08-13
Refactor the path type of com.island.ohara.kafka.connector.storage.Storage


eechih:
Clear path types can avoid confusion, so I prefer to use java.nio.file.Path instead of String type.

- Refactor the type of input parameter "path" to java.nio.file.Path in all methods
- Refactor the type of result list to java.nio.file.Path in the **list()** method

e.g.
- boolean exists(String path) -->  boolean exists(Path path)
- Iterator\<String> list(String path) --> Iterator\<Path> list(Path path)

chia7712:
@eechih 麻煩要考慮一個階段停一下先把完成的設計相關文件完成 否則後面會變成一次要寫很多文件會搞得自己很痛苦

eechih:
> 先把完成的設計相關文件完成

Okok

eechih:
redesigned by #2572, so close this issue.

PR #2269, opened 2019-08-13
#2267 Refactor result type of list() to java.nio.file.Path in Storage

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2267 )


chia7712:
The NIO Path is associated to local file system by default. Wrapping the ftp path to NIO path is a bit weird and dangerous. Perhaps you should add enough comment that all you want to address is to normalize the String to Path.

eechih:
> Perhaps you should add enough comment that all you want to address is to normalize the String to Path.

@chia7712 I have added some comments to remind users not to use the method associated with the file operation directly. e.g. Path#toFile()

Cloud you review again? thanks.

Issue #2158, opened 2019-08-06
Apply TopicKey to Stream APIs


chia7712:
The Stream APIs still use deprecated interface of Topic object ...

1. - [x] for 0.8 (#2282)
1. - [x] for 0.7 (#2306)

saivirtue:
Is there any plain to backport 0.7?
I think there are a lot of changes needed to be accomplish after topicKey apply to 0.7. Could we use a hadcode prefix to 0.7 version? I think it may reduce the task effort for frontend.

chia7712:
> Could we use a hadcode prefix to 0.7 version?

Let us fix the group for streamapp in 0.7. I don't think the hard code to fix group for streamapp is a easy way :(

chia7712:
@konekoya @wu87988622 We are going to file PR to enable streamapp to accept topic key in 0.7. Please take a look at #2306

konekoya:
I'll add a new task in #2209 to address this

PR #2306, opened 2019-08-15
#2158 Apply TopicKey to Stream APIs (backport to 0.7)

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2158)


ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
```diff
@@ -231,18 +228,21 @@ trait StreamCollie extends Collie[StreamClusterInfo, StreamCollie.ClusterCreator
   protected def postCreateCluster(clusterInfo: ClusterInfo, successfulContainers: Seq[ContainerInfo]): Unit = {
     //Default do nothing
   }
+
+  /**
+    * get the containers for specific broker cluster. This method is used to update the route.
```
saivirtue: Is "This method is used to update the route" not necessary?

ohara-streams/src/main/java/com/island/ohara/streams/ostream/LaunchImpl.java
```diff
@@ -77,15 +78,36 @@ public static void launchApplication(
 
                   OStream<Row> ostream =
                       OStream.builder()
-                          .appid(streamDefinitions.get(StreamDefinitions.NAME_DEFINITION.key()))
+                          // There are many tests which don't pass correct arguments to env, and the
+                          // origin design
+                          // lack of enough checks to avoid incorrect arguments. It needs to be
+                          // refactor but not now.
+                          // TODO: add enough checks for env variables ... by chia
```
saivirtue: should this TODO be backport also?

ohara-streams/src/test/java/com/island/ohara/streams/ostream/TestSimpleStreamCounter.java
```diff
@@ -34,6 +34,10 @@
 import org.junit.Before;
 import org.junit.Test;
 
+// TODO: the streamapp requires many arguments from env variables.
+// This tests do not care the rules required by streamapp.
+// Fortunately (or unfortunately), streamapp lacks of enough checks to variables so the
+// non-completed settings to streamapp works well in this test ... by chia
```
saivirtue: ditto

ohara-agent/src/main/scala/com/island/ohara/agent/StreamCollie.scala
chia7712: yep. let me fix this.

ohara-streams/src/main/java/com/island/ohara/streams/ostream/LaunchImpl.java
chia7712: TODO is ok to 0.7 :)

saivirtue:
let me fix the error tests.

chia7712:
@wu87988622 please merge it after @saivirtue complete this PR.

ohara-streams/src/main/java/com/island/ohara/streams/ostream/LaunchImpl.java
chia7712: The TODO is not good if we put it in the public document for the inactive branch since the document is a kind of guarantee. By contrast, the source code is not a public guarantee so the TODO is fine.

Issue #2246, opened 2019-08-12
JAR_KEY_DEFINITION should not be mapped to both "jar key" and "jar url"


chia7712:
Having another "key" does not burn out our server.

saivirtue:
Let me take over this issue

PR #2309, opened 2019-08-15
#2246 JAR_KEY_DEFINITION should not be mapped to different field

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2246)


saivirtue:
> a unrelated question. there are two way to pass "config" to streamapp container - env and main arguments. why we use both ways?

They are for different purpose. main argument is used to "dry run" ; env is used to set values.

chia7712:
> main argument is used to "dry run" ; env is used to set values.

What is the purpose of “dry run” to our ohara? Passing the arguments via various channels may obstruct us from controlling the settings (for example, decode and check).

saivirtue:
move the discussion thread to #2312

Issue #2244, opened 2019-08-12
StreamDefinitions should be immutable since it is a json object


chia7712:
As a serializable object, opening the door to modify the inner state is really, really, really dangerous ....

saivirtue:
Let me take over this issue too :)

saivirtue:
Let me remove the serializable of StreamDefinitions class

PR #2315, opened 2019-08-15
 #2244 StreamDefinitions should be immutable since it is a json object

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2244)


ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefUtils.java
```diff
@@ -265,76 +249,74 @@
...
+              StreamDefUtils.NODE_NAMES_DEFINITION,
+              StreamDefUtils.VERSION_DEFINITION,
+              StreamDefUtils.REVISION_DEFINITION,
+              StreamDefUtils.AUTHOR_DEFINITION,
+              StreamDefUtils.TAGS_DEFINITION,
+              StreamDefUtils.JAR_KEY_DEFINITION,
+              StreamDefUtils.JAR_INFO_DEFINITION)
+          .collect(Collectors.toList());
+
+  private List<SettingDef> configDefs = new ArrayList<>();
```
chia7712: The name "utils", normally, is a collection of (static) helper methods/members. Having the inner non-static variable is weird.

There are two behaviors for this class. first is to show the default definitions. seconds is to parse the value according to definitions. My two cents is that we can have two classes for them.

saivirtue: OK

saivirtue:
retry

docs/src/user_guide.rst
```diff
@@ -695,7 +695,7 @@ running any Ohara containers, you need to install Kubernets first.
 We’ll walk you through this process with a few k8s commands:
 
 Install distribute mode for Kubernetes
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
chia7712: the purpose of this change?

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
```diff
@@ -327,14 +67,15 @@ public StreamDefinitions add(SettingDef config) {
    */
   public StreamDefinitions addAll(List<SettingDef> configList) {
```
chia7712: If the setters are not necessary, making the StreamDefintitions immutable is always a good idea.

docs/src/user_guide.rst
saivirtue: The rst plugin of Intellij suggest that. It seems less warning in IDE but want hurt the doc compile. :)

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
saivirtue: Thanks for the suggestion, I will try to use immutable object instead.

docs/src/user_guide.rst
chia7712: Got it

saivirtue:
retry connector

ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefinitions.java
```diff
@@ -16,338 +16,80 @@
...
-          .key("streamMetricGroup")
-          .group(CORE_GROUP)
-          .orderInGroup(ORDER_COUNTER.getAndIncrement())
-          .optional("streamapp")
-          .readonly()
-          .internal()
-          .valueType(Type.STRING)
-          .build();
+  StreamDefinitions(List<SettingDef> configDefs) {
+    this.configDefs = configDefs;
```
chia7712: a nit: copy the input list and wrap it to a immutable list.

saivirtue: :)

Issue #2413, opened 2019-08-21
document the gradle commands used to build/test backend code only


chia7712:
My friend (@brandboat) complained that our "How to build" does not say how to build/test backend code only, and it causes that he had to install a lot of software to build Ohara Manager on his local.

vitojeng:
Trust me, I really understand his pain. XD

PR #2414, opened 2019-08-21
#2413 document the gradle commands used to build/test backend code only

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2413)


docs/src/how_to_build.rst
```diff
@@ -70,6 +70,23 @@ to simplify the development of ohara.
    you testing your code on `official QA`_ which offers the powerful
    machine and IT envs.
 
+**Run All UTs for backend only**
```
vitojeng: Now we can use **Definition List** if here don't need a header.
https://ohara.readthedocs.io/en/latest/contrib-docs/quick_example.html#definition-lists

By the way, could you also check in this file?

docs/src/how_to_build.rst
```diff
@@ -70,6 +70,23 @@ to simplify the development of ohara.
...
+
+.. code:: sh
```
vitojeng: Here we can use **console block**:
https://ohara.readthedocs.io/en/latest/contrib-docs/quick_example.html#console-block

docs/src/how_to_build.rst
```diff
@@ -70,6 +70,23 @@ to simplify the development of ohara.
...
+
+   gradle clean test -PskipManager
```
vitojeng: After we use **console** code-blolck, we can add `$` now:
> **$** gradle clean test -PskipManager

docs/src/how_to_build.rst
chia7712: There are some notes for different command, and I'm not sure whether the form "definition list" is able to accept "note" block. Hence, I keep the current style.

vitojeng: > I'm not sure whether the form "definition list" is able to accept "note" block.

Yes. Definition list accept "note" blcok.
You can reference the **docs/src/contrib-docs/quick_example.rst**. The whole quick example use definition list with many **block**, include code-block and note.

You also can build docs in the local for further checking.

docs/src/how_to_build.rst
konekoya: Does this mean we have to add the `$` whenever we use the code-block?

Issue #2319, opened 2019-08-15
Support default definition in CsvSink


eechih:
We want the connector that inherits from CsvSink to use the default definition.

PR #2455, opened 2019-08-26
#2319 support default definitions in CSV sink connector

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2319 )


ohara-kafka/src/main/java/com/island/ohara/kafka/connector/csv/CsvSinkConnector.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.stream.Stream;
+
+public abstract class CsvSinkConnector extends RowSinkConnector {
+
+  /**
+   * Define the configuration for the connector.
+   *
+   * @return The SettingDef for this connector.
+   */
+  protected List<SettingDef> _definitions2() {
```
chia7712: Could we have another name for this method?

eechih: I want to maintain the same format (starting with underline, the same rules as RowSinkConnector), and have same meaning as the original. So it is quite difficult. :(
Do you have any suggestion? @chia7712

chia7712: It is ok to remove the "final" from Row connector definition(), and then this new connector is able to override it to have _definition() be abstract.

eechih: Address your comment. Please review again, thanks. @chia7712

Issue #2481, opened 2019-08-28
Create worker tags value not saved


wu87988622:
Post
/v0/workers

Request payload
```
{
  "name": "wk3",
  "nodeNames": [
    "node00"
  ],
  "jarKeys": [],
  "jmxPort": 6082,
  "clientPort": 48522,
  "brokerClusterName": "jh99fgb4nr",
  "groupId": "vy1va6a77s",
  "configTopicName": "vw80h5jnk4",
  "offsetTopicName": "jtzeid2seg",
  "statusTopicName": "76glpp4998",
  "tags": {
    "zookeeper": {
      "name": "3t2h0li795",
      "imageNmae": "oharastream/zookeeper:0.8.0-SNAPSHOT"
    },
    "broker": {
      "name": "jh99fgb4nr",
      "imageNmae": "oharastream/broker:0.8.0-SNAPSHOT"
    }
  }
}
```
Response
```
{
  "name": "wk3",
  "nodeNames": [
    "node00"
  ],
  "jmxPort": 6082,
  "clientPort": 48522,
  "brokerClusterName": "jh99fgb4nr",
  "groupId": "vy1va6a77s",
  "configTopicName": "vw80h5jnk4",
  "offsetTopicName": "jtzeid2seg",
  "statusTopicName": "76glpp4998",
  "tags": {
    "zookeeper": {
      "name": "3t2h0li795",
      "imageNmae": "oharastream/zookeeper:0.8.0-SNAPSHOT"
    },
    "broker": {
      "name": "jh99fgb4nr",
      "imageNmae": "oharastream/broker:0.8.0-SNAPSHOT"
    }
  },
  "offsetTopicPartitions": 1,
  "lastModified": 1566976683531,
  "connectors": [],
  "jarInfos": [],
  "imageName": "oharastream/connect-worker:0.8.0-SNAPSHOT",
  "deadNodes": [],
  "statusTopicReplications": 1,
  "configTopicPartitions": 1,
  "offsetTopicReplications": 1,
  "configTopicReplications": 1,
  "statusTopicPartitions": 1
}
```
Put
/v0/workers/wk3/start

Response
`202`

But next time ...
Get
/v0/workers/wk3

Response
```
{
  "name": "wk3",
  "nodeNames": [
    "node00"
  ],
  "jmxPort": 6082,
  "clientPort": 48522,
  "brokerClusterName": "jh99fgb4nr",
  "groupId": "vy1va6a77s",
[...]

PR #2485, opened 2019-08-28
#2481 Create worker tags value not saved

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2481)


ohara-configurator/src/main/scala/com/island/ohara/configurator/route/package.scala
```diff
@@ -701,14 +701,21 @@ package object route {
...
       )
+      // TODO: in fact, this may be a useless action since all Getters do update state before generating response.
       .flatMap { cluster =>
-        // add the up-to-date object to store
-        // TODO: in fact, this may be a useless action since all Getters do update state before generating response.
-        store.add[Cluster](cluster)
+        // update the new state to object
+        // Note: there are some fields (ex: tags) are defined by user (api)
+        // we should remain the same values after update the data store
+        store.addIfPresent[Cluster](
```
chia7712: this is a embarrassed scenario. Should we recover the properties for cluster when there is no properties but cluster exists?

saivirtue: > Should we recover the properties for cluster when there is no properties but cluster exists?

Does this case exist?
For API, a cluster creation is along to another property, if there was no property exists, the creation should be failed.

chia7712: > Does this case exist?

Maybe. Who know all cases in production? At any rate, it should be worth of having comment to explain your view :)

saivirtue: Sure thing.

jackyoh:
retry

Issue #2229, opened 2019-08-12
Address unnecessary requests that are made in react hooks based API


konekoya:
@wu87988622

PR #2494, opened 2019-08-29
#2229 Address unnecessary requests that are made in react hooks based…

wu87988622:
… API

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2229 )

### DIFF
--- a/ohara-manager/client/src/components/controller/useFetchApi.js
+++ b/ohara-manager/client/src/components/controller/useFetchApi.js
@@ -24,29 +24,41 @@ const useFetchApi = url => {
   const { showMessage } = useSnackbar();
   const [response, setResponse] = useState();
   const [isLoading, setIsLoading] = useState(true);
-  const [refetchState, refetch] = useState(null);
-  const request = useCallback(
-    async url => {
-      try {
-        const res = await axiosInstance.get(url);
-        const isSuccess = get(res, 'data.isSuccess', false);
+  const [refetchState, refetch] = useState(true);
+  const request = useCallback(async () => {
+    try {
+      const res = await axiosInstance.get(url);
+      const isSuccess = get(res, 'data.isSuccess', false);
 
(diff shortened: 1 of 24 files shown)


ohara-manager/client/src/components/controller/useFetchApi.js
```diff
@@ -24,29 +24,41 @@ const useFetchApi = url => {
...
+      }
+    };
+    fetchData();
+
+    return () => {
+      didCancel = true;
+    };
+  }, [refetchState, request]);
 
   return { data: response, isLoading, refetch };
```
konekoya: Hmm, how about making the refetch like: `refetch: () => refetch(true)`

I didn't test this. But it should work okay

ohara-manager/client/src/components/pages/NodePage/NodeListPage.js
```diff
@@ -134,15 +134,15 @@ const NodeListPage = () => {
           isOpen={isNewModalOpen}
           handleClose={handleModalClose}
           handleConfirm={() => {
-            refetch();
+            refetch(true);
```
konekoya: With the above change, we won't need to pass the `true` here

konekoya:
@wu87988622 could you also take a look at the API tests? Since we're using tags now. The tests should also assert that `tags` field is included in the API response

wu87988622:
> @wu87988622 could you also take a look at the API tests? Since we're using tags now. The tests should also assert that `tags` field is included in the API response

OK I will add task in api test

wu87988622:
retry manager-e2e

wu87988622:
retry manager-e2e

konekoya:
retry manager-e2e

wu87988622:
retry manager-e2e

Issue #2438, opened 2019-08-22
Adapt new metadata api for all clusters


chia7712:
#2243 has redefined the meanings of nodeNames - it is a "setting" from user input. Hence, the nodeNames should be moved to "settings". This change breaks the APIs compatibility.

- [x] Ohara Manager should fetch the object data from settings (@wu87988622 @konekoya) (#2684)
- [x] move nodeNames to settings for StreamClusterInfo (#2521)
- [x] move nodeNames to settings for ZookeeperClusterInfo (#2552 )
- [x] move nodeNames to settings for BrokerClusterInfo (#2575)
- [x] move nodeNames to settings for WorkerClusterInfo (#2516)
- [x] remove stale fields for StreamClusterInfo (#2739)
- [x] remove stale fields for ZookeeperClusterInfo (#2740)
- [x] remove stale fields for BrokerClusterInfo (#2755)
- [x] remove stale fields for WorkerClusterInfo (#2756)
- [x] update docs to remove stale fields for clusters

~This issue is blocked by #1736 since most clusters don't have "settings" field yet :(~
Done

Issue #2549, opened 2019-09-04
fix flaky IT tests TestXXXStreamApp.testDeadNodes


saivirtue:
Reference Jenkins tasks: https://builds.is-land.com.tw/job/PreCommit-OHARA/5156/

TestXXXStreamApp.testDeadNodes
error message
```
java.lang.IllegalStateException: timeout
	at com.island.ohara.common.util.CommonUtils.await(CommonUtils.java:151)
	at com.island.ohara.common.util.CommonUtils.await(CommonUtils.java:119)
	at com.island.ohara.it.IntegrationTest$.await(IntegrationTest.scala:68)
	at com.island.ohara.it.IntegrationTest.await(IntegrationTest.scala:39)
	at com.island.ohara.it.agent.BasicTests4StreamApp.testDeadNodes(BasicTests4StreamApp.scala:409)
```

chia7712:
this may be duplicate to #2475.

chia7712:
BTW, there is a related issue (#2393)

saivirtue:
Got it. Let me edit the description.

saivirtue:
This issue is related to #2257 and should be resolved. Close this issue now.

PR #2531, opened 2019-09-03
#2191 Support settings field for zookeeper

saivirtue:
### Checklist
- [x] [Review Contribution Guidelines](https://github.com/oharastream/ohara/blob/master/CONTRIBUTING.md)
- [x] Add Reviewers (see [Authors](https://github.com/oharastream/ohara#authors))
- [x] Add Assignees (of course, you are the assignee by default!)
- [x] DON'T set a Milestone
- [x] DON'T add any labels (PR labels will be added automatically by Jenkins)
- [x] Link to an issue(You can do so by using the required by keyword: required by #2191)
- [x] Add unit tests (or please explain why this PR is not worth having new tests)
- [ ] Pass QA (Sometimes you may encounter an existing flaky test which would obstruct you from getting QA passed. If the failed tests are unrelated, you can add a comment in the PR thread)

PR #2552, opened 2019-09-04
#2438 move nodeNames to settings for ZookeeperClusterInfo

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2438)


saivirtue:
> We have to keep the compatibility for Ohara Manager, hence we have to add the nodeNames to json response of ZookeeperClusterInfo. (it means you have to add pair (nodeNames -> "xxx") to ZOOKEEPER_CLUSTER_INFO_JSON_FORMAT manually)

You mean [this line](https://github.com/oharastream/ohara/blob/81baeb572dacd286b00f1fe53eba0c3053e5b5c5/ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/ZookeeperApi.scala#L188)?

I had updated that in #2531

chia7712:
retry it

chia7712:
retry it

ohara-agent/src/main/scala/com/island/ohara/agent/ZookeeperCollie.scala
```diff
@@ -124,7 +124,6 @@ trait ZookeeperCollie extends Collie[ZookeeperClusterInfo] {
                         .nodeNames(successfulContainers.map(_.nodeName).toSet)
```
chia7712: my bad :( 

the nodeNames should be configured by input Creation than the sum of running containers. I have filed a issue for it (#2557)

saivirtue: That's OK.
I think #2557 is related to #2549 ; they both have same problem.
Let me reference the relation.

saivirtue:
The failed tests are not related to this issue, will merge later.

Issue #2276, opened 2019-08-13
Implement SMBStorage


eechih:
Implement SMBStorage

PR #2566, opened 2019-09-05
#2276 implement SmbFileSystem and IT test

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2276 )


ohara-client/src/main/scala/com/island/ohara/client/filesystem/smb/SmbClient.scala
```diff
@@ -0,0 +1,269 @@
...
+import com.hierynomus.smbj.share.DiskShare
+import com.hierynomus.smbj.{SMBClient, SmbConfig}
+import com.island.ohara.client.filesystem.{FileSystem, FileType}
+import com.island.ohara.common.annotations.Optional
+import com.island.ohara.common.exception.OharaSmbException
+import com.island.ohara.common.util.{CommonUtils, Releasable}
+
+import scala.collection.JavaConverters._
+
+trait SmbClient extends FileSystem
```
chia7712: Could we rename SmbClient to SmbFileSystem?

chia7712: Or we offer builder for smb than have a explicit class called SmbClient

chia7712:
what the difference between FileSystem and Storage?

eechih:
> what the difference between FileSystem and Storage?

The purpose of creating `FileSystem` trait is to standardize the client implementation of those file system, like `FtpClient`, `SmbClient` etc. So I will fire an issue to refactor `FtpClient`. 

It seems that `FileSystem` and `Storage` are very similar, so it is possible to keep only one of them in the future.

ohara-client/src/main/scala/com/island/ohara/client/filesystem/smb/SmbClient.scala
eechih: > Or we offer builder for smb than have a explicit class called SmbClient

We already have a builder method for creating `SmbClient`.

```
SmbClient
        .builder()
        .hostname(hostname.get)
        .port(Integer.parseInt(port.get))
        .user(user.get)
        .password(password.get)
        .shareName(shareName.get)
        .build)
```

chia7712: I meant we don't have a class called SmbClient.

For example, FileSystem.smbBuilder

chia7712:
> It seems that FileSystem and Storage are very similar, so it is possible to keep only one of them in the future.

It would be better to integrate them into a single class if they are similar ...

ohara-client/src/main/scala/com/island/ohara/client/filesystem/smb/SmbClient.scala
eechih: Oh~ I misunderstood, I will fix it. Thanks. :)

eechih: @chia7712 address your comment, please review again, thanks.

eechih:
> It would be better to integrate them into a single class if they are similar ...

This task will be done in another issue #2572 

@chia7712 Thanks you for your assistance.

jackyoh:
retry connector

jackyoh:
retry it

eechih:
retry otherIT

eechih:
retry otherIT

Issue #2570, opened 2019-09-05
Apply group concept for all cluster


saivirtue:
Currently, we already live in a group world, this rule should apply to clusters.
After this issue, the actual container name format will be:
`{PREFIX_KEY}-{GROUP}-{CLUSTER_NAME}-{SERVICE}-{HASH}`

Both GROUP and CLUSTER_NAME length limit is 20 chars for now.

- [x] Apply group concept for zookeeper (#2583)
- [x] Apply group concept for broker (#2624)
- [x] Apply group concept for worker (#2629)
  - Remove stale method
- [x] Apply group concept for stream (#2602)
- [x] Refactor collie.creator infra to benefit the api.request (see [comment](https://github.com/oharastream/ohara/pull/2624#pullrequestreview-286617999)) (#2633)

chia7712:
```scala
  private[this] def updateState[Cluster <: ClusterInfo: ClassTag](cluster: Cluster, metricsKey: Option[String])(
    implicit meterCache: MeterCache,
    collie: Collie[Cluster],
    executionContext: ExecutionContext): Future[Cluster] =
    collie
      .clusters()
      .map(
        _.keys
          .find(_.name == cluster.name)  // _.key == cluster.key
```
@saivirtue should we make above change in this issue?

saivirtue:
>  should we make the above change in this issue?

We should :)

PR #2624, opened 2019-09-11
 #2570 Apply group concept for broker

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2570)


ohara-agent/src/main/scala/com/island/ohara/agent/BrokerCollie.scala
```diff
@@ -382,9 +379,9 @@ object BrokerCollie {
 
     override def create(): Future[BrokerClusterInfo] = {
       // initial the basic creation required parameters (defined in ClusterInfo) for broker
-      //TODO : it is ok that we don't add the group() since client will auto fill default value for us
-      //TODO: add group() in #2570 for broker
-      val creation = request.name(clusterName).imageName(imageName).nodeNames(nodeNames).creation
+      // Note: We add all the fields in collie and they may not required in client api
```
chia7712: nice comment!

saivirtue:
> seems we are able to make collie.creator depend on Api.request. If so, the fields in basic creator can be removed and all setter can be abstract. The sub classes should have their way to set nodeNames, group, name and so on. The benefit is that the creator get pure interface and we can get rid of weird helper method (for example, doCopy).

This needs a sub-task to finish. Already added in #2570

saivirtue:
retry it

saivirtue:
retry it

PR #2628, opened 2019-09-11
#2625 Remove nodename in hostname label of k8s

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2625)


saivirtue:
retry configurator

saivirtue:
retry it

saivirtue:
retry it

saivirtue:
retry sshCollie

saivirtue:
retry sshConfigurator

saivirtue:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/ZookeeperCollie.scala
```diff
@@ -111,8 +111,7 @@ trait ZookeeperCollie extends Collie[ZookeeperClusterInfo] {
                         // container env quickly. Also, the specific string enable us to pick up the "true" settings
                         // from envs since there are many system-defined settings in container envs.
                           + toEnvString(creation.settings),
-                        // zookeeper doesn't have advertised hostname/port so we assign the "docker host" directly
-                        hostname = node.name
+                        hostname = containerName
```
chia7712: The hostname exposed by ZK is what the hostname does be. It means the  containerName will be exposed to all ZK client, and it causes problem in resolving dns since not all client be aware of “container name”. By contrast, the node name, which is a true node in production, is familiar with all clients, and the resolving should work well.

saivirtue: > it causes problem in resolving dns since not all client be aware of “container name”. 

That is awkward that k8s did the job for us, and if we need such a solution to resolve dns problem, I will need to add more comment about this case.
Let me revert this change.

chia7712: The issue I concerned does happen outside the clusters and there is no k8s dns. For example, we have to connect to zk to dig in something weird and assigning “container name” to hostname, which is exposed by zk, will cause trouble to us. By contrast, assigning “host” name to container hostname can simplify our life.

saivirtue: > we have to connect to zk to dig in something weird and assigning “container name” to hostname, which is exposed by zk, will cause trouble to us.

Good example, let me add in.

saivirtue:
retry it

saivirtue:
retry otherIT

saivirtue:
retry it

Issue #2682, opened 2019-09-18
Increase the name and group field length limit for more flexibility


saivirtue:
We use (group, name) as our object key in Ohara, and the most strict rule is cluster object:
1. name & group : [0-9], [a-z], [-]
2. name & group length <= 20

This is because we have to assemble a container name in clusters with the following format:
`{occl|k8soccl}_{$group}_{$name}_{zk|bk|wk|stream}_{HASH_VALUE}`

And k8s has a "resource name maximum length of 253 characters" restriction.

To simplify the rules and let the group be more flexibility to accept the "hierarchy-group" usage,
this issue will aim to the following tasks:

1. - [x] the sum of name and group length should less or equal than 100 in clusters (#2701)
    - name & group : [0-9], [a-z]
1. - [x] apply name and group rules to objects of UI used (#2722)
    - connector
    - pipeline
    - topic
    - node (only restrict length < 63)
1. - [x] fix the illegal object name in IT tests (@chia7712) (#2804)

Also see [Ohara Stream Key Restrictions](https://docs.google.com/document/d/1UwzXUuGsKKez9loAo6dcGiZU2P3FfvOV6olP4vmYTpk/edit?usp=sharing)

chia7712:
The length limit to name is 60 as well?

saivirtue:
> The length limit to name is 60 as well?

Only group part. Name field could be part of "group name" in some cases, so I prefer remain the name length <= 20:
For example, 
workspace: group=default, name=abc
pipeline: group=abc-mypipeline, name=mypipeline

chia7712:
So rule easing is for UI? The group is made of “group” and name from other service?

If so, I prefer to limit the sum of “group length” and “name length”. The story of inconsistent limit(for group and name) is a bit mysterious to backend. BTW, using our json refiner to implement the limit should be ok.

saivirtue:
> So rule easing is for UI? The group is made of “group” and name from other service?

Any rule here is to unify all restrictions to one. For API, we still need to handle duplicated name problems:
For example, we have two workspaces:
{"group": "abc", "name": "wk01"}, {"group": "abc", "name": "wk02"}
and
wk01 has two pipelines: p01, p02
- p01 contains connectors: c01, c02
wk02 has one pipeline: p01
- p01 contains connector: c02

The above samples could be accomplished by API or by UI, and should be successful.
However, the group for connector here should use the format `wk_name+pipeline_name` to avoid possible duplication

[...]

PR #2701, opened 2019-09-19
#2682 the sum of name and group length should less or equal than 100 in clusters

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2682)


ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/JsonRefiner.scala
```diff
@@ -300,6 +319,17 @@ trait JsonRefiner[T] {
...
+  /**
+    * add your custom check for a set of keys' values.
+    * Noted: We don't guarantee the order of each (key, value) pair since this method is intend
+    * to do an "aggregation" check, like sum or length check.
+    *
+    * @param keys keys
+    * @param checkers checkers
+    * @return this refiner
+    */
+  def allValuesChecker(keys: Set[String], checkers: Set[JsValue] => Unit): JsonRefiner[T]
```
chia7712: what happen if the keys of input request don't match the "keys" completely?

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/JsonRefiner.scala
```diff
@@ -615,6 +651,11 @@ object JsonRefiner {
           // 5) check the keys
           keyCheckers.foreach(_(fields.keySet))
 
+          // 6) check the values of specific keys
+          allvaluesCheckers.foreach {
+            case (keys, f) => f(fields.filter(field => keys.contains(field._1)).values.toSet)
```
chia7712: toSet de-duplicates the values. For example:

```json
{
  "name": "a",
  "group": "a"
}
```

the stuff passed via allvaluesCheckers will be "a" than "a", "a"

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/JsonRefiner.scala
saivirtue: Neglect in my first design. But requireKeys should be more suitable.

konekoya:
retry

saivirtue:
retry configurator

Issue #2708, opened 2019-09-19
Verify the version of all official connectors


eechih:
As title

PR #2716, opened 2019-09-19
#2708 Verify the version of all official connectors

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2708 )


ohara-connector/src/test/scala/com/island/ohara/connector/TestConnectors.scala
```diff
@@ -0,0 +1,63 @@
...
+      definitions.foreach { d =>
+        val key = d.key()
+        if (key == "version" || key == "revision" || key == "author")
+          if (d.defaultValue() == "unknown")
+            throw new AssertionError(s"The $connName $key cannot be unknown.")
+      }
+    }
+  }
+
+  private[connector] def officialSourceConnectorClasses: Seq[Class[_ <: RowSourceConnector]] = {
```
chia7712: please deliberate on integrating those two methods.

ohara-connector/src/test/scala/com/island/ohara/connector/TestConnectors.scala
```diff
@@ -0,0 +1,63 @@
...
+    val reflections = new Reflections()
+    val classes = reflections.getSubTypesOf(classOf[RowSourceConnector]).asScala
+    classes
+    // the abstract class is not instantiable.
+      .filterNot(clz => Modifier.isAbstract(clz.getModifiers))
+      .filter(clz => clz.getName.startsWith("com.island.ohara.connector"))
```
chia7712: Is this filter necessary? the connectors found out in test scope should be owned by ohara.

eechih: > Is this filter necessary?

Yes, because some connectors must be avoided, such as com.island.ohara.kafka.connector.DumbSource

ohara-connector/src/test/scala/com/island/ohara/connector/TestConnectors.scala
eechih: mm... I need to think about it.

konekoya:
retry manager-api

ohara-connector/src/test/scala/com/island/ohara/connector/TestConnectors.scala
eechih: I've updated the PR, could you give it a look again?

Issue #2731, opened 2019-09-21
Cluster APIs should honor cluster key rather than cluster name


chia7712:
cluster accept group now but our APIs still require user to pass name rather than key...

ObjectKey's json parser is about to accept following json representation.
```json
"key": {
  "group":"g",
  "name":"n"
}
```
```json
"key": {
  "name":"n"
}
```
```json
"key": "n"
```
Hence, the "default group" to all objects should be identical to make a global json parser for key type.

The sub issues are listed below.

1. - [x] unify default group for all objects (#2754)
1. - [x] enable json parser of "key" to accept following type: {"group": "g", "name": "n"}, {"name":"n"} and "n" (#2764)
1. - [x] Broker APIs should honor cluster key (#2808)
1. - [x] remove zookeeperClusterName from Broker APIs (#2845)
1. - [x] Worker APIs should honor cluster key (#2817)
1. - [x] remove brokerClusterName from Worker APIs
1. - [x] Stream APIs should honor cluster key (#2800)
      ~- [ ] Stream does not handle the stale filed in definition (#2831)~
1. - [x] remove brokerClusterName from Stream APIs (#2834)
1. - [x] Collie should honor cluster key (#2839)

related to #2769

saivirtue:
cluster accept creation without group specified. If we use key instead of group and name, how could we create an object with default group?

chia7712:
> If we use key instead of group and name, how could we create an object with default group?

The Scala APIs accept the key and the json APIs take both key and name.

The former is for consistent in writing pure scala code. Converting key to name is a weird and ugly style across our project. 

The later take “key” but accept the key lacking of “group”. For example: the following requests are identical.

bkKey: {
group:default,
name:”a”
}

bkkey: a

saivirtue:
> The Scala APIs accept the key and the json APIs take both key and name.

Since scala APIs are part of "client", this conclusion is very important for later developers/users.
Could you add this conclusion to the `key()` comment? :)

chia7712:
> Could you add this conclusion to the key() comment? :)

sure, boss

saivirtue:
/cc @oharastream/frontend 

The latest changes of cluster APIs are traced here, I will remove all the stale fields in this release (0.8).
If you had any problem, please let me know :)

chia7712:
> The latest changes of cluster APIs are traced here, I will remove all the stale fields in this release (0.8).
If you had any problem, please let me know :)

[...]

PR #2754, opened 2019-09-23
#2731 unify default group for all objects

chia7712:
### Checklist
- [x] DON'T set a Milestone
refactor ~


chia7712:
retry it

ohara-client/src/main/scala/com/island/ohara/client/configurator/v0/package.scala
```diff
@@ -30,7 +30,7 @@ package object v0 {
     * This field is protected since it is a "default" to all APIs scopes than a "global" default to whole project.
     * However, each APIs scope may have custom default value of group.
     */
-  private[v0] val GROUP_DEFAULT: String = "default"
+  val GROUP_DEFAULT: String = "default"
```
saivirtue: This change violates the description above...
What's the original purpose to use different "default" group for each APIs?

chia7712: > What's the original purpose to use different "default" group for each APIs?

some objects have more meaningful "default group". For example, the default group of connector should be equal to worker cluster name as it makes sense that we can't deploy two connectors which have same name on the same worker cluster. Topic also has similar story.

However, we didn't adopt above mechanism since it complicate our APIs.

saivirtue: > However, we didn't adopt above mechanism since it complicate our APIs.

Got it.

By the way, I think we should remove the description about the custom part:
> /**
    * the default group to all objects.
    * the group is useful to Ohara Manager. However, in simple case, the group is a bit noisy so we offer the default group to all objects when
    * input group is ignored.
    *
    * This field is protected since it is a "default" to all APIs scopes than a "global" default to whole project.
    * However, each APIs scope may have custom default value of group.
    */

chia7712: roger that

chia7712:
retry configurator

chia7712:
retry configurator

chia7712:
retry configurator

PR #2764, opened 2019-09-24
#2731 enable json parser of key to accept following type: {"group": "…

chia7712:
…g", "name": "n"}, {"name":"n"} and "n"

### Checklist
- [x] DON'T set a Milestone


ohara-client/src/test/scala/com/island/ohara/client/configurator/v0/TestKeyApi.scala
```diff
@@ -87,4 +87,28 @@ class TestKeyApi extends OharaTest with Matchers {
...
+    key.name() shouldBe name
+  }
+
+  @Test
+  def testPureStringInConnectorKey(): Unit = {
+    val name = CommonUtils.randomString()
+    val key = CONNECTOR_KEY_FORMAT.read(JsString(name))
+    key.group() shouldBe GROUP_DEFAULT
+    key.name() shouldBe name
+  }
```
saivirtue: Could we add the tests for `{"name": "n", "group": "g"}` and `pure string` also?

chia7712: > {"name": "n", "group": "g"}

ok

> pure string

pure string can't be parsed to json object. I don't want to add a testing-purpose class so I just create a JsString to test :)

saivirtue: So, the tests here is for the json type
```json
"key": {
  "n"
}
```
right?

Since the title elaborates json parser should accept three different types, I thought we will have three different type tests 😕

chia7712: > "key": {

this string is converted to JsObject rather than JsString so it is not what we want to test.

chia7712:
unrelated error

Issue #2766, opened 2019-09-24
Add default value to the keys 'task.total', 'task.hash' of CsvSourceConfig


eechih:
as title.
- task.total 's default value is 1
- task.hash 's default value is 0

chia7712:
Personally, it is a bit dangerous to have default value for them since both of them must be defined correctly to avoid race condition.

eechih:
> Personally, it is a bit dangerous to have default value for them since both of them must be defined correctly to avoid race condition.

+1, I will revert this PR #2767

Issue #2775, opened 2019-09-25
Cannot to start smb source connector because some keys cannot to set during the startup phase.


eechih:
Cannot to start smb source connector because some keys "task.total", "task.hash" cannot to set during the startup phase.

PR #2776, opened 2019-09-25
Revert "#2766 Add default value to the keys 'task.total', 'task.hash' of CsvSourceConfig"

eechih:
revert this PR because it is a bit dangerous to have default value for them since both of them must be defined correctly to avoid race condition.

ref: #2767


chia7712:
@eechih there is nothing change?

eechih:
> there is nothing change?

@chia7712 I will take other solution. ref #2775

eechih:
> there is nothing change?

@chia7712 Sorry! I made a mistake in the revert. I have already created another PR #2783

PR #2800, opened 2019-09-26
#2731 Stream APIs should honor cluster key

saivirtue:
### Checklist
- [x] DON'T set a Milestone


saivirtue:
retry it

chia7712:
the error is caused by #2682

saivirtue:
> the error is caused by #2682

How so? Do the IT tests use illegal chars for object key?

chia7712:
> How so? Do the IT tests use illegal chars for object key?

upper case :(

but, there is other error in our IT. let me dig in it :)

saivirtue:
> but, there is other error in our IT. let me dig in it :)

OK. For more reliable PR, I will wait for the problem resolved and retry the it tests.

chia7712:
retry it

chia7712:
retry it

chia7712:
retry it

chia7712:
retry otherIT

saivirtue:
The failed tests are traced in #2818 and #2821, will merge this PR later

PR #2831, opened 2019-09-27
#2731 Stream does not handle the stale filed in definition

saivirtue:
### Checklist
- [x] DON'T set a Milestone


ohara-streams/src/main/java/com/island/ohara/streams/config/StreamDefUtils.java
```diff
@@ -37,6 +37,18 @@
   /** This is the default configurations we will load into {@code StreamDefUtils}. */
   private static final AtomicInteger ORDER_COUNTER = new AtomicInteger(0);
 
+  // TODO: remove this stale field (see https://github.com/oharastream/ohara/issues/2731)
+  public static final SettingDef BROKER_CLUSTER_NAME_DEFINITION =
```
chia7712: A required and internal field means server have to "define" it. Does Configurator handle it ?

saivirtue: It is a workaround to handle the stale "brokerClusterName" field check. user could add the stale field in "settings", we should remain this stale definition to let StreamRoute checking pass in staring phase.

saivirtue:
Since stream is using dynamic definitions and settings, the compatibility should not be problem. Will close this issue.

Issue #2886, opened 2019-10-03
Remove column definition from JsonIn and JsonOut


chia7712:
the column is useless to JsonIn and JsonOut. Also, showing it on UI may confuse user to enter meaningless configs.

1. - [x] for master (#2887)
1. - [x] for 0.8 (#2890)

PR #2890, opened 2019-10-03
#2886 Remove column definition from JsonIn and JsonOut

chia7712:
backport to 0.8 (related to #2886)


chia7712:
retry sshConnector

Issue #2938, opened 2019-10-07
JsonIn should use java.BigDecimal rather than scala.BigDecimal


chia7712:
Although we don't long for consistence of all connectors' output, this little change brings the tremendous benefit for connectors implemented by java.

related to #2937

1. - [x] for 0.9 (#2939)
1. - [x] for 0.8 (#2940)

PR #2940, opened 2019-10-07
#2938 JsonIn should use java.BigDecimal rather than scala.BigDecimal

chia7712:
backport to 0.8 (#2938)


chia7712:
retry configurator

chia7712:
retry configurator

chia7712:
retry configurator

Issue #3013, opened 2019-10-15
k8s collie does not define the correct host name


chia7712:
```
java.net.UnknownHostException: 63eaab37c-zk-280ca70: Name or service not known
	at java.net.Inet4AddressImpl.lookupAllHostAddr(Native Method)
	at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:929)
	at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1324)
	at java.net.InetAddress.getAllByName0(InetAddress.java:1277)
	at java.net.InetAddress.getAllByName(InetAddress.java:1193)
	at java.net.InetAddress.getAllByName(InetAddress.java:1127)
	at java.net.InetAddress.getByName(InetAddress.java:1077)
	at org.apache.zookeeper.server.quorum.QuorumPeer$QuorumServer.recreateSocketAddresses(QuorumPeer.java:180)
	at org.apache.zookeeper.server.quorum.QuorumPeer$QuorumServer.<init>(QuorumPeer.java:152)
	at org.apache.zookeeper.server.quorum.QuorumPeerConfig.parseProperties(QuorumPeerConfig.java:240)
	at org.apache.zookeeper.server.quorum.QuorumPeerConfig.parse(QuorumPeerConfig.java:152)
	at org.apache.zookeeper.server.quorum.QuorumPeerMain.initializeAndRun(QuorumPeerMain.java:104)
	at org.apache.zookeeper.server.quorum.QuorumPeerMain.main(QuorumPeerMain.java:81)
```

```scala
      .nodeName(containerInfo.nodeName)
      // this hostname has a length limit that <=63
      // see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
      // zookeeper doesn't have advertised hostname/port so we assign the "docker host" directly
      // Note: We should assign "node" name to the container hostname directly here to avoid some
      // dns problem. For example, we may want to connect to zk to dig something issue and assign
      // node name here can save our life to solve the connection problem...
      .hostname(node.name)
```

the above code is related to old solution and we have, now, changed the format of server.x for all zk services. 

caused by #2970

PR #3014, opened 2019-10-15
#3013 k8s collie does not define the correct host name

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3013)


ohara-agent/src/main/scala/com/island/ohara/agent/k8s/K8SZookeeperCollieImpl.scala
```diff
@@ -42,13 +42,7 @@ private class K8SZookeeperCollieImpl(node: NodeCollie, k8sClient: K8SClient)
...
         containerInfo.portMappings.flatMap(_.portPairs).map(pair => pair.hostPort -> pair.containerPort).toMap)
       .nodeName(containerInfo.nodeName)
-      // this hostname has a length limit that <=63
-      // see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
-      // zookeeper doesn't have advertised hostname/port so we assign the "docker host" directly
-      // Note: We should assign "node" name to the container hostname directly here to avoid some
-      // dns problem. For example, we may want to connect to zk to dig something issue and assign
-      // node name here can save our life to solve the connection problem...
-      .hostname(node.name)
+      .hostname(containerInfo.hostname)
```
saivirtue: Since `containerInfo.hostname` is different with `node.name` now, we should leave some "comment" to remind mistake usage.

ohara-it/src/test/scala/com/island/ohara/it/agent/BasicTests4Collie.scala
```diff
@@ -240,7 +240,7 @@ abstract class BasicTests4Collie extends IntegrationTest with Matchers {
     log.info(s"verify number of log... done")
     result(zk_logs(clusterKey)).foreach(log =>
       withClue(log) {
-        log.contains("exception") shouldBe false
+        log.toLowerCase().contains("exception") shouldBe false
```
saivirtue: could we add some tests to check the container.nodename is part of container.name also?

ohara-agent/src/main/scala/com/island/ohara/agent/k8s/K8SZookeeperCollieImpl.scala
chia7712: got it

ohara-it/src/test/scala/com/island/ohara/it/agent/BasicTests4Collie.scala
chia7712: roger that

jackyoh:
retry otherIT

chia7712:
retry k8sConnector

Issue #2994, opened 2019-10-14
[Manager] Scaffolding new frontend codebase


konekoya:
- [x] Remove old UI components from manager (#3025)
- [x] Create Mui color theme for the new UI (#3039)
- [x] Update manager's development guideline (#3045)
- [x] Add Ohara brand and main navigation (#3066)
- [x] Add snackbar context into the new UI (#3078)
- [x] Add an E2E test case to guard against 404 page (#3085)
- [x] Display a workspace automatically if there's an existing one (#3096)
- [x] Create context APIs for accessing shared global data (I've created a few contexts in this issue, more context can be added if necessary)

 ### Things I don't know what/how to do yet:
- [x] Active state for workspace
- [x] Pick another icon for the event log

konekoya:
@wu87988622 @saivirtue

konekoya:
I'm thinking about completely ditching frontend unit test since:
1. It's hard to write and maintain these tests
2. We often spend MUCH more time in the unit tests rather than the actual feature...
3. E2E tests give us way more confidence than unit tests 
4. We tend to write tests that don't bring much value for us

PR #3078, opened 2019-10-25
#2994 Add snackbar context into the new UI

konekoya:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #2994)


ohara-manager/client/src/App.js
```diff
@@ -38,6 +39,12 @@ const Main = styled.main`
 `;
 
 const App = () => {
+  const showMessage = useSnackbar();
```
konekoya: Since `showMessage` is the only method that we're going to be using in the code, I simply return a `function` instead of an `object`

ohara-manager/client/src/App.js
```diff
@@ -38,6 +39,12 @@ const Main = styled.main`
...
+
+  useEffect(() => {
+    showMessage('Hello world!');
```
konekoya: We're implementing this with a `useReduce` and the function is actually a `dispatch`, so it won't cause trouble like we had in the old UI (infinite loop)

ohara-manager/client/src/App.js
```diff
@@ -48,7 +55,9 @@ const App = () => {
             <Route
               exact
               path="/"
-              component={() => <h1>Hello ohara manager!</h1>}
+              component={() => (
+                <h1 onClick={() => showMessage('Ah?')}>Hello ohara manager!</h1>
```
konekoya: Make sure the message can also be displayed within an event handler

ohara-manager/client/src/context/SnackbarContext.js
```diff
@@ -0,0 +1,109 @@
...
+    const { type, message } = action;
+    switch (type) {
+      case 'SHOW':
+        return { isOpen: true, message };
+
+      case 'HIDE':
+        return { isOpen: false, message: '' };
+
+      default: {
+        throw new Error(`Unhandled action type: ${type}`);
```
konekoya: This is helpful for debugging when the given type is not valid

konekoya:
retry manager-e2e

Issue #3093, opened 2019-10-27
[Manager] Refactor tests of /v0/info API to adapt new structure


chia7712:
#3091 has added /v0/info/configurator to replace /v0/info, and this issue aims to complete the API migration.

- [x] add tests for zookeeper, broker, worker and configurator (#3110)
- [x] add tests for stream (move to #3028)

chia7712:
@wu87988622 @konekoya @saivirtue I noticed that /v0/info is already in APIs test so #3091 did not break the APIs.

konekoya:
@wu87988622 @konekoya @saivirtue I noticed that /v0/info is already in APIs test so #3091 did not break the APIs.

I think we can adapt the new API now as it's not used in the UI yet and the API tests are being rewritten, cc @saivirtue

saivirtue:
> @wu87988622 @konekoya @saivirtue I noticed that /v0/info is already in APIs test so #3091 did not break the APIs.

Please notice me after the #3091 closed.

chia7712:
> Please noticing me after the #3091 closed.

ping @saivirtue  :)

saivirtue:
All the relative tasks are done in #3028, close this issue now.

PR #3110, opened 2019-10-29
#3093 add tests for zookeeper, broker, worker and configurator

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3093)


chia7712:
@saivirtue could you also remove the deprecated route in InfoRoute? You are absolutely capable to modify the backend code :)

ohara-manager/client/cypress/api/info.test.js
```diff
@@ -14,22 +14,65 @@
...
 
+/* eslint-disable no-unused-expressions */
+
+import * as infoApi from '../../src/api/infoApi';
+
 describe('Info API', () => {
-  it('fetchInfo', () => {
-    cy.fetchInfo().then(response => {
+  it('fetchConfiguratorInfo', async () => {
+    await infoApi.fetchConfiguratorInfo().then(response => {
```
konekoya: I'm not sure if `Async Await` works in Cypress or not, but we should choose only one pattern and still to it for the sake of consistency: So in this case, use either `Async Await` (if it works okay here) or `Promise` :)

ohara-manager/client/cypress/api/info.test.js
```diff
@@ -14,22 +14,65 @@
...
         data: { isSuccess, result },
       } = response;
 
       const { mode, versionInfo } = result;
-      const { date, revision, version } = versionInfo;
+      const { branch, version, user, revision, date } = versionInfo;
 
       expect(isSuccess).to.eq(true);
 
+      // we using fake configurator to test ; the value should be "FAKE"
```
konekoya:
```suggestion
      // we're using fake configurator in our tests; so this value should be "FAKE"
```

ohara-manager/client/cypress/api/info.test.js
saivirtue: OK

ohara-manager/client/cypress/api/info.test.js
```diff
@@ -14,22 +14,65 @@
...
+      expect(imageName).to.be.a('string');
+      expect(imageName).to.include(serviceName);
+
+      expect(definitions).to.be.an('array');
+      expect(definitions.length > 0).to.be.true;
+
+      const objectKeyDefinitions = definitions
+        .map(definition => definition.key)
+        .filter(key => key === 'group' || key === 'name');
+      expect(objectKeyDefinitions.length).to.eq(2);
```
konekoya: Looks like you're explicitly testing these two keys here?

saivirtue: Yep.
We cannot assert other "definitions" here, but each service should have it's own `objectKey`, i.e, {name, group}

saivirtue:
retry

ohara-manager/client/cypress/api/info.test.js
konekoya: As per our discussion, we should fetch the defs first, then add some fake data and save these configs in the backend and assert the response :)

saivirtue:
retry manager-api

chia7712:
retry manager-api

konekoya:
retry it/sshConnector

konekoya:
retry sshConnector

Issue #3170, opened 2019-11-03
[Manager] Provide better faling test error from API test


wu87988622:
Making it easier to troubleshoot and consume from CI  

- [x]  Extend type assertion of API tests to be more informative (#3263)
- ~[ ] using console.log() or console.warn() to print the tests progress~
It is not a good solution for now since cypress has the following issues are not resolved
-- https://github.com/cypress-io/cypress/issues/448
-- https://github.com/cypress-io/cypress/issues/300
-- https://github.com/cypress-io/cypress/issues/4245
- ~[ ] collect the console log to file~

konekoya:
@wu87988622 I recall we have talked about this on the other day. We can print out the log (console.error or console.warn emitting from Cypress) while running the test. And after the test is done, collect the log can write to a file which can be used by Jenkins later

saivirtue:
As the description says, we don't have a good solution to print out the console log. Close this issue now.

PR #3263, opened 2019-11-13
#3170 Extend type assertion of API tests to be more informative

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3170)


saivirtue:
/cc @oharastream/frontend 
I try to add some messages of each assertion for better error tracing.
Is that OK for you guys?
I will finish the rest API tests in this PR.

chia7712:
Could we see both response and request via this PR?

saivirtue:
> Could we see both response and request via this PR?

It's hard to say. Since the major purpose of this PR is to add more message of each "expect", i.e., the log or console won't get print out during QA.
I want to add the req/res data either, but it should be done in another task of #3170

konekoya:
> I try to add some messages of each assertion for better error tracing.
Is that OK for you guys?
I will finish the rest API tests in this PR.

The only concern here is it's going to be quite verbose when writing these tests. If Cypress (or mocha) provides a better assert error output, then we won't have to add this manually...

I will try to whip up an example and file an issue over Cypress' repo and see if they can help us out :)

ohara-manager/client/cypress/configs/cypress.api.json
```diff
@@ -3,7 +3,8 @@
   "integrationFolder": "cypress/api",
   "reporterOptions": {
     "mochaFile": "../test-reports/apiReport-[hash].xml",
-    "toConsole": true
+    "toConsole": true,
+    "jenkinsMode": true
```
saivirtue: from [mocha-junit-reporter](https://github.com/michaelleeallen/mocha-junit-reporter#full-configuration-options):
> if set to truthy value will return xml that will display nice results in Jenkins

konekoya: @saivirtue Does this config affect the output report?

saivirtue: The most obvious changes are the `name` and `className`:

**before**
```xml
<testcase name="Broker API createBroker" time="0.897" classname="createBroker">
```

**after**
```xml
<testcase name="createBroker" time="4.481" classname="Broker API">
```

saivirtue: But it seems not to break our Jenkins behavior, I thought we could add this config?
ping @chia7712

Issue #3317, opened 2019-11-19
add "sinceSeconds" to Log APIs to fetch logs since relative time


saivirtue:
As title.

For now, it seems we could query full logs only for all services?

For example,
`GET /logs/zookeeper/{NAME}?group=abc`

saivirtue:
the related wireframe is here:
https://zpl.io/29zQvdY

chia7712:
It is necessary to UI so ... YES .. Configurator should offer this function :)

chia7712:
> Could you link the related issue to this one?

let me address it in this one :)

PR #3319, opened 2019-11-19
#3317 add "sinceSeconds" to Log APIs to fetch logs since relative time

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3317)


chia7712:
retry

ohara-agent/src/main/scala/com/island/ohara/agent/docker/DockerClient.scala
```diff
@@ -273,6 +280,12 @@ object DockerClient {
         .map(msg => if (msg.contains("ERROR:")) throw new IllegalArgumentException(msg) else msg)
         .getOrElse(throw new IllegalArgumentException(s"no response from ${agent.hostname}"))
 
+    override def log(name: String, sinceSeconds: Long): String =
+      agent
+        .execute(s"docker container logs $name --since=${sinceSeconds}s")
```
saivirtue: It seems the implementation of docker has different behavior:
> --since    Show logs since timestamp (e.g. 2013-01-02T13:23:37) or relative (e.g. 42m for 42 minutes)

If we use `?sinceSeconds=10`:
1. docker: will output the logs from **1970/1/1T00:00:10** to now
2. k8s: will output the the logs in the time range [ current() - 10s , current() ]

If we use `?sinceSeconds=10s`:
1. docker: will output the logs in the time range [ current() - 10s , current() ]
2. k8s: will output the the logs in the time range [ current() - 10s , current() ]

chia7712: > If we use ?sinceSeconds=10:
docker: will output the logs from 1970/1/1T00:00:10 to now

I had tested the docker option on my local. I didn't get the result what you described. The result is indeed in the time range.

saivirtue: > I had tested the docker option on my local.

I missing the postfix "s" in the docker logs command :(

chia7712: @saivirtue thanks for the confirmation :)

chia7712:
retry otherIT

chia7712:
retry otherIT

chia7712:
retry sshConnector

chia7712:
retry sshConfigurator

chia7712:
unrelated error

Issue #3087, opened 2019-10-25
Implement shabondi source routes


vitojeng:
required by #3022

vitojeng:
reopne for minor refactor

PR #3359, opened 2019-11-22
#3087 Implement Shabondi source routes

vitojeng:
required by #3087 

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3087)


ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
```diff
@@ -0,0 +1,120 @@
...
+class Config(raw: Map[String, String]) {
+  import DefaultDefinitions._
+  import scala.collection.JavaConverters._
+
+  def serverType: String = raw(SERVER_TYPE_KEY)
+  def port: Int          = raw(PORT_KEY).toInt
+
+  def sourceToTopics: Seq[TopicKey] = TopicKey.toTopicKeys(raw(SOURCE_TO_TOPICS_KEY)).asScala
+
+  def sinkFromTopic: TopicKey = TopicKey.toTopicKey(raw(SINK_FROM_TOPIC_KEY))
```
chia7712: Why the output is single way ?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
```diff
@@ -0,0 +1,120 @@
...
+    .reference(SettingDef.Reference.TOPIC)
+    .displayName("Source topic")
+    .documentation("The topic that Shabondi will pull rows from")
+    .optional(Type.OBJECT_KEYS)
+    .build
+    .registerDefault
+
+  implicit private class RegisterSettingDef(settingDef: SettingDef) {
+    def registerDefault: SettingDef = {
+      defaultDefinitions + (settingDef.key -> settingDef)
```
chia7712: "+" creates a new map so you should use "+="

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
```diff
@@ -16,24 +16,65 @@
...
+  private val producer = KafkaClient.producer(config.brokers)
+
+  def route(topicKeys: Seq[TopicKey]): Route = {
+    implicit val jsonStreamingSupport: JsonEntityStreamingSupport = EntityStreamingSupport.json()
+    (post & path("v0")) {
+      handleExceptions(exceptionHandler) {
+        entity(as[RowData]) { rowData =>
+          val row = JsonSupport.toRow(JsObject(rowData))
+
+          complete((StatusCodes.OK, sendRowFuture(topicKeys, row).map(_ => "success")))
```
chia7712: Given that it is a "true" data flow via restful APIs, should we make response after completing the sending? Or we should process the input data in the background?

chia7712: BTW, the http code "OK" is enough. The message "success" is a bit duplicate

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
```diff
@@ -0,0 +1,89 @@
...
+      .builder()
+      .keySerializer(Serializer.ROW)
+      .valueSerializer(Serializer.BYTES)
+      .offsetFromBegin()
+      .topicName(topicName)
+      .connectionProps(brokers)
+      .build()
+
+  def sendRow(producer: Producer[Row, Array[Byte]], topicName: String, row: Row): Future[RecordMetadata] = {
+    assert(producer != null)
```
chia7712: why not using scala future chain to avoid the multiple null check ?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
```diff
@@ -0,0 +1,89 @@
...
+    FutureConverters.toScala(sender.send().asInstanceOf[CompletableFuture[RecordMetadata]])
+  }
+
+  def pollOnce(
+    brokers: String,
+    topicName: String,
+    timeoutSecond: Long,
+    expectedSize: Int
+  ): Seq[Consumer.Record[Row, Array[Byte]]] = {
+    val c = consumer(brokers, topicName)
```
chia7712: kafka consumer is a heavy component so it would be better to reuse it :)

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
```diff
@@ -16,24 +16,65 @@
...
+    extends Directives {
+  import JsonSupport._
+  import actorSystem.dispatcher
+
+  implicit private val timeout: Timeout = Timeout(3 seconds)
+
+  private lazy val log = Logging(actorSystem, classOf[SourceRoute])
+
+  private val exceptionHandler = ExceptionHandler {
+    case ex: Exception =>
```
chia7712: Throwable is more suitable

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
vitojeng: Do you mean that user can consume multiple topics through single Shabondi sink instance?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
vitojeng: Thanks a lot!

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
chia7712: yep. all components in pipeline should support topics-to-topics

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
vitojeng: > Given that it is a "true" data flow via restful APIs, should we make response after completing the sending? Or we should process the input data in the background?

You are right. Shabondi needs to be able to handle special situations more elegantly.
For example, we should give a error response to user if kafka client give a exception in call back function. 
There should have a lot detail work here in the future, I feel.

vitojeng: > BTW, the http code "OK" is enough. The message "success" is a bit duplicate

Ok, thanks!

chia7712: > There should have a lot detail work here in the future, I feel.

got it

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
vitojeng: This just used in unit test for simple case. 
It should move to testing code. :)

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/WebServer.scala
```diff
@@ -16,17 +16,120 @@
...
+  }
+}
+
+/**
+  * reference: akka.http.scaladsl.server.HttpApp
+  */
+private abstract class AbstractWebServer extends Directives {
+  private val serverBinding = new AtomicReference[ServerBinding]()
+
+  protected val actorSystemRef = new AtomicReference[ActorSystem]()
```
chia7712: why to keep a ref for actor?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
vitojeng: Got it.

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
vitojeng: This line only check for producer.
Not sure your mean, can you provide an example?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
```diff
@@ -0,0 +1,89 @@
...
+      .keySerializer(Serializer.ROW)
+      .valueSerializer(Serializer.BYTES)
+      .offsetFromBegin()
+      .topicName(topicName)
+      .connectionProps(brokers)
+      .build()
+
+  def sendRow(producer: Producer[Row, Array[Byte]], topicName: String, row: Row): Future[RecordMetadata] = {
+    assert(producer != null)
+    val sender = producer.sender().key(row).value(KafkaClient.emptyValue).topicName(topicName)
```
chia7712: you don't need to fill empty array for value. ignoring it is ok to our producer.

vitojeng: Got it

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
vitojeng: Got it

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
```diff
@@ -0,0 +1,120 @@
...
+    .key(SERVER_TYPE_KEY)
+    .group(coreGroup)
+    .orderInGroup(orderNumber)
+    .optional(SERVER_TYPE_SOURCE, Set(SERVER_TYPE_SOURCE, SERVER_TYPE_SINK).asJava)
+    .displayName("Shabondi server type")
+    .documentation("The server type when Shabondi service start.")
+    .build
+    .registerDefault
+
+  val PORT_KEY = "shabondi.port"
```
chia7712: "shabondi.client.port" is better since it is straightforward when we start to add more xxx.port to shabondi (for example, shabondi.jmx.port)

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
chia7712: > This line only check for producer.

the check means there is (possibly) a "way" to input null producer. We should change the code style to eliminate the "possibility"

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
vitojeng: good point!

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Utils.scala
vitojeng: Got it

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/WebServer.scala
vitojeng: actorsystem ref passed from sub-class, we need keep the ref of actorsystem in AbstractWebServer for use in Future(bindingFuture) callback method

chia7712:
retry sshConnector

chia7712:
retry sshConfigurator

chia7712:
retry agent

chia7712:
unrelated error

PR #3416, opened 2019-11-28
#3087 Shabondi minor refactor

vitojeng:
required by #3087 

### Checklist
- [x] DON'T set a Milestone


ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
```diff
@@ -35,7 +35,8 @@ object SourceRoute {
 }
 
 class SourceRoute(config: Config)(implicit val actorSystem: ActorSystem, implicit val materializer: ActorMaterializer)
-    extends Directives {
+    extends Directives
+    with KafkaSupport {
```
chia7712: Mixing the trait to a *Public* class is a bit dangerous to ohara since we have dynamically check (ex, code style check, method check). the mixing implicitly adds Public methods to the Public class and it is hard to observe it.

ohara-shabondi/src/test/scala/com/island/ohara/shabondi/TestStreamGraph.scala
```diff
@@ -0,0 +1,66 @@
...
+      // Send row to two topics
+      val future = StreamGraph.fromSendRow(producer, topics, row).run()
+
+      Await.result(future, Duration.Inf)
+
+      // assertion
+      val rowsTopic1: Seq[Consumer.Record[Row, Array[Byte]]] =
+        pollTopicOnce(brokerProps, topicKey1.name(), 10, columnSize)
+      rowsTopic1.size should ===(1)
+      rowsTopic1(0).key.get.cells.size should ===(columnSize)
```
jackyoh: === 應該和 shouldBe 是一樣的意思

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
vitojeng: yeah, you are right.  most of classes in shabondi package should be private, IMO.
does this would be better?

ohara-shabondi/src/test/scala/com/island/ohara/shabondi/TestStreamGraph.scala
vitojeng: 嗯...幾乎一樣 :)
matcher 寫法太多...最後也不知怎麼地就習慣全都用 === 了

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SourceRoute.scala
chia7712: Narrowing the scope is always better to project. the thing I worried is only that the "extends xxxTrait" brings a lot of methods "attached" to the class, and they are visible to all. 

At any rate, this is just code style issue. I'm ok with this PR :)

vitojeng:
not related error

Issue #3180, opened 2019-11-04
[Manager] Create 404, 500 and some other error routes and pages


konekoya:
- [x] Set up 404 and 500 routes and error pages (#3375)
- [x] Create 501 error page (#3425)

PR #3425, opened 2019-11-29
#3180 Create 501 error page

konekoya:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3180)


ohara-manager/client/src/api/inspectApi.js
```diff
@@ -70,6 +72,11 @@ export const getStreamsInfo = async (params = {}) => {
   return fetchServiceInfo(kind.stream, params);
 };
 
+export const getManagerInfo = async () => {
+  const res = await axiosInstance.get(`${url}/${kind.manager}`);
+  return responseUtil(res, inspectManagerBody);
+};
+
```
saivirtue: hmm...we need a new test case for this new member:)

konekoya: Right, didn't know about that. I will add one :)

bin/start-service.sh
```diff
@@ -39,7 +39,7 @@ done
 
 if [ "$service" == "manager" ]; then
   cd "$PROJECT_HOME/manager"
-  exec node ./start.js $ARGS
+  exec NODE_ENV=production node ./start.js $ARGS
```
konekoya: @chia7712 This needs your review :)
Or is there a way to test this piece of code locally? Manager now needs this logic to determine whether the code is running in dev mode or prod mode so it can consume the correct paths for ohara_verison

chia7712: Could we use “export a=b” to set the env variables? Also, we expose envs via ohara-env.sh.

konekoya: No problem. Should I rebuild manager image in order to test this out?

chia7712: It is ok to test this PR on your local. We can make hot-fix if this PR does make something broken.

konekoya: Updated the PR :)

ohara-manager/client/src/api/inspectApi.js
konekoya: Test added

Issue #3378, opened 2019-11-25
Implement shabondi sink routes


vitojeng:
required by #3022

PR #3489, opened 2019-12-09
#3378 Shabondi sink routes(part 1)

vitojeng:
required by #3378

### Checklist
- [x] DON'T set a Milestone


ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
```diff
@@ -101,6 +119,17 @@ object DefaultDefinitions {
...
     .registerDefault
 
+  val SINK_MODE = "shabondi.sink.mode"
+  val SINK_MODE_DEFINITION = SettingDef.builder
+    .key(SINK_MODE)
+    .group(coreGroup)
+    .orderInGroup(orderNumber)
+    .optional(SINK_MODE_EXCLUSIVE, SINK_MODES.asJava)
+    .displayName("Sink mode")
+    .documentation("...")
```
vitojeng: @chia7712 Have any suggestion about the documentation? :)

chia7712: IIRC, Sink has a field - (consumer) group - to detect the “mode“.

With the same group, all data are exclusive. By contrast, defining the different group can fetch the data which are taken by other requests.

In short, the field “mode” is useless to shabondi sink :)

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/RowQueueProducer.scala
```diff
@@ -0,0 +1,112 @@
...
+            log.trace("  queue size: {}, row size: {}", queue.size, rows.size)
+          }
+        } else {
+          log.trace("  paused. queue size: {}", queue.size)
+        }
+        Thread.sleep(50)
+      } // while
+    } finally {
+      consumer.close()
+      log.debug("RowProducer stopped.")
```
chia7712: RowProducer?

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/Config.scala
vitojeng: You are right. I totally forgot it!! 😢 
Will remove it.

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SinkRouteHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+private[shabondi] class SinkRouteHandler(config: Config) extends RouteHandler {
+  import Boot._
+  import JsonSupport._
+
+  private val log = Logging(actorSystem, classOf[SinkRouteHandler])
+
+  // TODO:  We should provide a
+  private val threadPool =
+    Executors.newFixedThreadPool(4, new ThreadFactoryBuilder().setNameFormat("shabondi-sinkpool-%d").build())
+  private val queue                = new LinkedBlockingQueue[Row]()
```
chia7712: We have limit, traverse and size requests to the blocking queue. Hence, ArrayBlockingQueue is more suitable to this case.

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SinkRouteHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+      log.error(ex, ex.getMessage)
+      complete((StatusCodes.InternalServerError, ex.getMessage))
+  }
+
+  private def fullyPollQueue(queue: BlockingQueue[Row]): Seq[Row] = {
+    if (queue.isEmpty)
+      Seq.empty[Row]
+    else {
+      val buffer = ArrayBuffer.empty[Row]
+      while (queue.size() > 0) {
```
chia7712: There is no transaction covering both size and poll functions. A more safe code style is to check the returned value from poll. Add to buffer and keep the loop if the value is NOT null. By contrast, break the loop

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/RowQueueProducer.scala
vitojeng: should be `RowQueueProducer`, will update it.

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SinkRouteHandler.scala
vitojeng: Sorry for some misunderstanding here.
The `limitSize` in class RowQueueProducer is not real limit size for the queue.
Because the poll size of consumer will possibility exceed the `limitSize`, so I choice LinkedBlockingQueue.

ohara-shabondi/src/main/scala/com/island/ohara/shabondi/SinkRouteHandler.scala
vitojeng: Got it.

vitojeng:
retry clientIT

vitojeng:
retry collieIT

vitojeng:
retry connectorIT

Issue #3466, opened 2019-12-06
Make necessary adjustments to definitions API in order to meet the UI requirements


wu87988622:
1. - [x] add an new flag "updatable" to definition to protect the value (@chia7712) (#3470)
1. - [x] Display console warning of API request in development environment only (@saivirtue )(#3692)
1. - [x] Change "GROUP"'s internal to true so it won't display in the UI @wu87988622 #3556 
1. - [x] Change "ZOOKEEPER_CLUSTER", "BROKER_CLUSTER", and "WORKER_CLUSTER"'s  internal to true so they won't display in the UI @wu87988622  #3556 
1. - [x] Add new flag "updatable" and regular to manager #3564 
1. - [x] Disable user to enter value in blacklist and show the recommended values to user (#3167) #3564

chia7712:
@oharastream/frontend I have opened another issue to merge "editable" and "updatable" (see #3480). It is going to break the compatibility so please ping me if you are ready to be in new APIs world :)

saivirtue:
All subtasks were done, close this issue now.

PR #3564, opened 2019-12-18
#3466 Add new flag "updatable" and regular to manager

wu87988622:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3466)

### DIFF
--- a/ohara-manager/client/src/utils/validate.js
+++ b/ohara-manager/client/src/utils/validate.js
@@ -46,7 +46,7 @@ export const maxNumber = max => value => {
 
 export const validWithRegex = regex => value => {
   return regex.test(value)
-    ? 'The value does not meet the rules' + regex
+    ? `The value does not meet the rules ${regex}`
     : undefined;
 };
 
@@ -63,7 +63,7 @@ export const validWithDef = def => value => {
   }
   if (blacklist.length > 0) {
     if (blacklist.find(black => black === value)) {
-      return 'This value is in the blacklist';
+      return 'The value is listed in the blacklist and so cannot be used';
     }
   }
   if (regex) {
--- a/ohara-manager/client/src/utils/validate.js
+++ b/ohara-manager/client/src/utils/validate.js
@@ -44,5 +44,34 @@ export const maxNumber = max => value => {
   return value <= max ? undefined : `The value must be less than ${max}`;
 };
 
+export const validWithRegex = regex => value => {
+  return regex.test(value)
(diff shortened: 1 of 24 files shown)


ohara-manager/client/src/utils/validate.js
```diff
@@ -44,5 +44,34 @@ export const maxNumber = max => value => {
...
+  const { necessary, regex, blacklist } = def;
+  if (necessary === 'REQUIRED') {
+    const requiredValue = required(value);
+    if (requiredValue) {
+      return requiredValue;
+    }
+  }
+  if (blacklist.length > 0) {
+    if (blacklist.find(black => black === value)) {
+      return 'This value is in the blacklist';
```
konekoya:
```suggestion
      return 'The value is listed in the blacklist and so cannot be used';
```

ohara-manager/client/src/utils/validate.js
```diff
@@ -44,5 +44,34 @@ export const maxNumber = max => value => {
...
+    }
+  }
+  if (regex) {
+    const regex = validWithRegex(def.regex);
+    const regexValue = regex(value);
+    if (regexValue) {
+      return regexValue;
+    }
+  }
+  return undefined;
```
konekoya: I'd return `null` here since that's more specific. A function without a "return" keyword which returns `undefined`

ohara-manager/client/src/utils/validate.js
```diff
@@ -44,5 +44,34 @@ export const maxNumber = max => value => {
   return value <= max ? undefined : `The value must be less than ${max}`;
 };
 
+export const validWithRegex = regex => value => {
+  return regex.test(value)
+    ? 'The value does not meet the rules' + regex
```
konekoya: Shouldn't there be some whitespace?  `The value does not meet the rules ' + regex`

Or simply use a template string `The value does not meet the rules ${regex}`

wu87988622:
> Just be mindful that using min/max in an input does not completely block users from entering values that are greater/lesser than the min/max values :)

I will create this issue later processing

ohara-manager/client/src/utils/validate.js
wu87988622: ok

ohara-manager/client/src/utils/validate.js
wu87988622: ok

Issue #3658, opened 2019-12-31
[Manager] Fix plugins - save to the right target


eechih:
as title

konekoya:
@eechih this can be closed by now, right :)

PR #3659, opened 2019-12-31
#3658 [Manager] fix plugins - save to the right target

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3658 )


ohara-manager/client/src/context/api/brokerApi.js
```diff
@@ -106,14 +94,13 @@ export const createApi = context => {
...
-          throw new Error(res.title);
-        }
+        const ensuredValues = { ...values, group };
+        const res = await brokerApi.update(ensuredValues);
+        if (res.errors) throw new Error(res.title);
+
         const data = generateClusterResponse({ values: res.data });
         showMessage(res.title);
-        return data;
+        return { ...data, ...key(ensuredValues) };
```
konekoya: I think adding a verb for this key function would be easier to read, like:
```js
getKey()
generateKey()
createKey()
```
Just some examples, it should depend on your use case here :)

eechih: No problem! I will use getKey() :)

PR #3762, opened 2020-01-08
#3760 Add another required() method with recommanded values into SettingDef

vitojeng:
required by #3760 

### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3760)


ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
```diff
@@ -722,6 +722,18 @@ public Builder required(Type valueType) {
       return checkAndSet(valueType, Necessary.REQUIRED, null);
     }
 
+    public Builder required(Set<String> recommendedValues) {
+      this.recommendedValues = Objects.requireNonNull(recommendedValues);
+      return checkAndSet(
+              Type.STRING, Necessary.REQUIRED, null);
+    }
+
+    public Builder required(String defaultValue, Set<String> recommendedValues) {
```
chia7712: it is NOT required if there is default value (and it is duplicate to https://github.com/oharastream/ohara/blob/master/ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java#L889)

ohara-configurator/src/main/scala/com/island/ohara/configurator/route/ObjectChecker.scala
```diff
@@ -43,7 +43,7 @@ import scala.reflect.ClassTag
   * unify all resource checks and produces unified error message.
   */
 trait ObjectChecker {
-  def checkList: ChickList
+  def checkList: CheckList
```
chia7712: good :)

ohara-common/src/main/java/com/island/ohara/common/setting/SettingDef.java
vitojeng: Got it

vitojeng:
retry manager-api

vitojeng:
retry collieIT

vitojeng:
retry connectorIT

Issue #3752, opened 2020-01-08
[Manager] Add default workspace name in quick mode


saivirtue:
Add a default workspace name for user to create workspace more quickly.

rules:
1. prefix: "quickworkspace" + serial number
2. if duplicated, serial number +1

related to #3739

- [x] v0.10.0 (#3773 )
- [x] v0.9.0 (#3774 )

eechih:
@saivirtue I will assign this issue to me :)

saivirtue:
@eechih 
You can keep going on this issue since #3746 has been merged :)

PR #3773, opened 2020-01-08
#3752 [Manager] Add default workspace name in quick mode

eechih:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3752 )


ohara-manager/client/src/components/Workspace/Create/hooks.js
```diff
@@ -0,0 +1,30 @@
...
+
+import { useMemo } from 'react';
+import { some } from 'lodash';
+
+import { useWorkspace } from 'context';
+
+export const useUniqueName = (prefix = 'quickworkspace') => {
+  const { workspaces = [] } = useWorkspace();
+  return useMemo(() => {
+    for (var postfix = 1; postfix <= Number.MAX_SAFE_INTEGER; postfix++) {
```
konekoya: We should probably use `let` instead of `var` here. `MAX_SAFE_INTEGER` is good trick, by the way :)

konekoya: We should add a lint rule for vars though...

Issue #3785, opened 2020-01-09
the run-time information of creating node is incorrect - it is always unavailable and resources are empty


chia7712:
from https://github.com/oharastream/ohara/issues/3706#issuecomment-571862957

1. - [x] for 0.10.0 (#3790)
1. - [x] for 0.9.0 (#3791)

saivirtue:
It is due to async problem. Let me take over this issue.

chia7712:
> It is due to async problem. Let me take over this issue.

the APIs don't generate correct runtime information as well. Please add extra task to this issue if you are going to fix something for UI :)

PR #3791, opened 2020-01-09
#3785 the run-time information of creating node is incorrect - it is …

chia7712:
backport to 0.9.x (related to #3785)


chia7712:
unrelated error

Issue #3694, opened 2020-01-06
Refactor components (including remove redundant "settings" field)


saivirtue:
Since the work of refactoring has a lot of changes that may cause the UI flow break, I move this issue here independently.

- [x] Replace api with context actions in components (#3849)
- [x] Refactor DevToolDialog component to use reducer structure (#3835)
~- [ ] Refactor cell selected action to event trigger~ #3813 had done this, nothing to do

PR #3849, opened 2020-01-17
#3694 Replace api with context actions in components

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3694)


ohara-manager/client/src/components/Workspace/Create/WorkspaceQuick.js
```diff
@@ -135,23 +127,10 @@ const WorkspaceQuick = props => {
...
-
-  const onDrop = async (file, values) => {
-    const parentKey = { group: GROUP.workspace, name: values.workspaceName };
-    const result = await fileApi.create({
-      group: hashByGroupAndName(GROUP.workspace, values.workspaceName),
-      file: file[0],
-      tags: { parentKey },
-    });
+  const onDrop = async file => {
+    const result = await createFile(file[0]);
```
eechih: @saivirtue  在 Create workspace 的階段下，ApiContext fileApi 的 fileGroup 會是 null 或指到上一個 workspace，所以直接使用 fileApi 應該會出現錯誤... 🤔

ohara-manager/client/src/components/Workspace/WorkspaceList.js
```diff
@@ -59,19 +61,25 @@ Statistic.propTypes = {
...
     const fetchTopics = async () => {
       const result = await topicApi.getAll();
       setTopics(result.errors ? [] : result.data);
     };
-    if (topics !== null) return;
-    fetchTopics();
-  }, [topics]);
+    // if we add some topic into current workspace
+    // re-fetch all topics again
+    if (topics === null || prevTopics !== currentTopics) {
```
konekoya: No a bit deal, but we could move this check to the top of this effect and so we can return early if these conditions are met :)

konekoya: E.g.,

```
React.useEffect(() => {

if (...conditions) return;

const fetchTopics = async => { ... function body }


// rest of our logic

}, [deps...])
```

ohara-manager/client/src/components/Workspace/Create/WorkspaceQuick.js
saivirtue: You are right :(
After all, we still have some exceptions that context action could not be used.

ohara-manager/client/src/components/Workspace/WorkspaceList.js
saivirtue: These checks are already in the top of useEffect :)
```javascript
React.useEffect(() => {
   const fetchTopics = ....;
   if(conditions) {
      fetchTopics();
   }
});
```

konekoya: Oops, I must have missed them

Issue #3161, opened 2019-11-01
[Manager] Create E2E tests for the new UI


konekoya:
- [x] Add tests to verify devTool dialog works well (@saivirtue) (#3868)
- [x] Add a test to verify toolbox of pipeline can be worked normally (#3960)
- Add tests to verify pipeline can be worked normally
  - [x] Add a test to verify links of connector and topic can be worked normally (#4033)
  ~~- [ ] Add a test to verify links of stream, connector and topic can be worked normally (#4045)~~
  ~~- [ ] Add a test to verify auto add topic can be worked normally~~
  ~~- [ ] Add tests to verify settings and actions can be worked normally~~
- [x] Add a test to verify all routes are working as expected (#3891)
- [x] Add a test to verify a workspace can be created via quick mode (#3935)
- [x] Add a test to verify the node dialog (#3942)
- [x] Add a test to verify the topic dialog (#3976)
- [x] Add a test to verify the files dialog (#3981)
- [x] Add a test to verify the eventLog dialog (@saivirtue ) (#4042)
- [x] Bumping cypress to the newest version (#3936)

Additional tests:
~~- [ ] Add a test to verify select cells could link to the correct position of DevTool dialog~~

saivirtue:
@konekoya 
I have pushed the in-progress commit of the first task into my repo (saivirtue/fix_3161_1)
feel free to pull it :)

konekoya:
@saivirtue would you mind adding a task for bumping cypress version to the newest version?

saivirtue:
> @saivirtue would you mind adding a task for bumping cypress version to the newest version?

Sure thing :)

konekoya:
> Sure thing :)

Thanks :)

saivirtue:
/cc @eechih 
Could you help me to accomplish the eventLog testing part?

eechih:
> Could you help me to accomplish the eventLog testing part?

Sure

PR #3942, opened 2020-02-04
#3161 Add a test to verify the node dialog

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #3161)


ohara-manager/client/cypress/e2e/nodeDialog.test.js
```diff
@@ -0,0 +1,295 @@
...
+
+    // press "ESC" back to node list
+    cy.get('body:visible').trigger('keydown', { keyCode: 27, which: 27 });
+    // press "ESC" again back to home page
+    cy.findAllByText(/^nodes$/i)
+      .first()
+      .trigger('keydown', { keyCode: 27, which: 27 });
+
+    cy.findByTitle('Create a new workspace').click();
+    cy.findByText(/^quick\sstart$/i)
```
konekoya: Hmm,  but now selector like this is a bit hard to read. I wonder if we can get rid of the escape char... 🤔

ohara-manager/client/cypress/e2e/nodeDialog.test.js
```diff
@@ -0,0 +1,295 @@
...
+
+    // Create a workspace for this node
+    // Step1: workspace name (using default)
+    cy.findAllByText(/^next$/i)
+      .filter(':visible')
+      .click();
+
+    // Step2: select nodes
+    cy.contains('button', 'Select nodes').click();
+    cy.findByText(new RegExp(`^${nodeHost}$`, 'i'))
```
konekoya: Also this

ohara-manager/client/cypress/e2e/nodeDialog.test.js
saivirtue: I think that's why I use ignore-case in some place only...It's hard to read in some situations.

ohara-manager/client/cypress/e2e/nodeDialog.test.js
saivirtue: Maybe variable cases could use case-sensitive only?

saivirtue:
retry

saivirtue:
retry manager

saivirtue:
retry manager-e2e

Issue #4007, opened 2020-02-12
Encourage all to use gradlew rather than gradle


chia7712:
inspired by #4005

the gradlew can resolve the version issue.

PR #4008, opened 2020-02-12
#4007 Encourage all to use gradlew rather than gradle

chia7712:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #4007)
I will test all dockerfiles locally 


gradlew
```diff
@@ -0,0 +1,188 @@
...
+    ;;
+  MINGW* )
+    msys=true
+    ;;
+  NONSTOP* )
+    nonstop=true
+    ;;
+esac
+
+CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
```
vitojeng: It seems we need append **gradle-wrapper.jar** ?

chia7712: gradle-wrapper.jar  is binary file. The gradlew script download it automatically if gradle-wrapper.jar  does not exist.

vitojeng: Where is the downalod script in gradlew script file? I cannot find it.

jackyoh: 下載的位址在 gradle/wrapper/gradle-wrapper.properties 的 distributionUrl 變數裡

vitojeng: yeah, you are right. But I think not contain any download script in the **gradlew** file.

chia7712: seems the downlaod is included by gradle-wrapper.jar :(

let me add the binary to our source

chia7712:
unrelated error

Issue #4046, opened 2020-02-17
 Integrate Shabondi with Configurator, part 2


vitojeng:
required by #3022 
follow up #3591 

Todo itmes:
- [x] Split boot class and definitions into two groups: source and sink(#4057)
- [x] Shabondi minor refactor(#4475)
- [x] Add shabondi info to InspectAPI
  - [x] Add shabondi info(#4461)
  - [x] Add running shabondi info(#4494) 
- [x] Dependency check of Shabondi with Broker and Topic(#4501)
- [x] Enable Shabondi JMX port and HEAP_OPTS(#4511)
- [x] Enable shabondi metrics
  - [x] Enable shabondi metrics(#4536)
  - [x] Enable shabondi metrics in configurator route(#4547)
- [ ] ~Add more testing~ (Move to #4559)

PR #4057, opened 2020-02-18
#4046 Split Shabondi boot class and definition into source and sink groups

vitojeng:
required by #4046 

### Checklist
- [x] DON'T set a Milestone


ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
```diff
@@ -0,0 +1,259 @@
...
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+sealed abstract class ShabondiType(val name: String, val className: String)
+
+final case object ShabondiSource extends ShabondiType("source", SourceBoot.getClass.getName.replaceAll("\\$", ""))
+
+final case object ShabondiSink extends ShabondiType("sink", SinkBoot.getClass.getName.replaceAll("\\$", ""))
+
+final object ShabondiType {
```
chia7712: see oharastream.ohara.client.Enum

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
```diff
@@ -0,0 +1,259 @@
...
+    .registerTo(basicDefinitionMap)
+
+  val IMAGE_NAME_DEFINITION = SettingDef.builder
+    .group(CORE_GROUP)
+    .key("imageName")
+    .orderInGroup(orderInGroup())
+    .displayName("Image name")
+    .documentation("The image name of this Shabondi running with")
+    .optional(IMAGE_NAME_DEFAULT)
+    .permission(SettingDef.Permission.READ_ONLY) // In manager, user cannot change the image name
```
chia7712: So we disable user to run different version of shabondi at runtime?

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
```diff
@@ -0,0 +1,259 @@
...
+    .key("shabondi.brokers")
+    .orderInGroup(orderInGroup())
+    .required(Type.STRING)
+    .displayName("Broker list")
+    .documentation("The broker list of current workspace")
+    .internal
+    .build
+    .registerTo(basicDefinitionMap)
+
+  val ROUTES_DEFINITION = SettingDef.builder
```
chia7712: Are there extra routes used by shabondi? The routes to services (for example, broker) are configured by collie, and this filed is used to accept custom routes passed by users.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
```diff
@@ -0,0 +1,259 @@
...
+    .key("revision")
+    .orderInGroup(orderInGroup())
+    .displayName("Revision")
+    .documentation("Revision of stream")
+    .optional(VersionUtils.REVISION)
+    .permission(SettingDef.Permission.READ_ONLY)
+    .build
+    .registerTo(basicDefinitionMap)
+
+  val AUTHOR_DEFINITION: SettingDef = SettingDef.builder
```
chia7712: version, revision and author are moved to common group. Also oharastream.ohara.common.setting.WithDefinitions has many helper methods to generate those common definitions.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
vitojeng: Got it.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
vitojeng: The same permission setting with Stream.
I guess we would enable it in the future?

chia7712: > I guess we would enable it in the future?

absolutely. Could you leave a issue link for it?

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
vitojeng: The extra route should not used by shabondi now. 
Will remove it.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
vitojeng: sure

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/ShabondiDefinitions.scala
vitojeng: Got it, thanks for remind.

Issue #4073, opened 2020-02-20
change the group of version, revision and author from "common" to "meta"


chia7712:
the common is the default group to all definitions so it is not suitable to the version, revision and author.

1. - [x]  for 0.10.0 (#4074)
1. - [x]  for 0.9.0 (#4081)

PR #4081, opened 2020-02-20
#4073 change the group of version, revision and author from "common" …

chia7712:
backport to 0.9.x

related to #4073


chia7712:
retry collieIT

Issue #4092, opened 2020-02-21
Display more details of error in parsing column type


chia7712:
this is inspired by #4079

1. - [x]  for 0.10.0 (#4093)
1. - [x]  for 0.9.0 (#4096)

PR #4096, opened 2020-02-21
#4092 Display more details of error in parsing column type

chia7712:
backport to 0.9.0 (related to #4092)


chia7712:
retry connectorIT

Issue #4342, opened 2020-03-17
[Manager] Adjust typescript project structure


saivirtue:
Base on [TypeScript Deep Dive](https://basarat.gitbook.io/typescript/), we have following things to do:
- [x] re-compile in runtime both .js and .ts source code file (#4344)
- [x] remove duplicated Typescript library in client (#4376)
- [x] Use enum with static functions to refine the project
  - Type of SettingDef in RenderDefinition

PR #4344, opened 2020-03-17
#4342 auto-compile both .js and .ts source code file

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #4342)
  - Only change json file


saivirtue:
retry manager-e2e

ohara-manager/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "preinstall": "node ./scripts/checkYarnVersion.js",
     "setup": "yarn install --frozen-lockfile && yarn --cwd client setup",
-    "start": "cross-env NODE_ENV=development nodemon start.js --exec \"yarn lint:server && node\"",
+    "start": "cross-env NODE_ENV=development nodemon start.js --watch 'client/*' --exec \"yarn lint:server && ts-node\"",
```
konekoya: So this excludes watching file changes made in ohara-manager, right?

ohara-manager/package.json
```diff
@@ -54,6 +54,8 @@
     "lodash": "4.17.15",
     "morgan": "1.9.1",
     "sha.js": "2.4.11",
+    "ts-node": "8.6.2",
+    "typescript": "3.8.3",
```
konekoya: Should we remove the typescript from client and just use this one for both ohara-manager and its client (UI)?

ohara-manager/package.json
saivirtue: Yep. We should watch client files only, right?

ohara-manager/package.json
saivirtue: Good point. Thanks for pointing out.

ohara-manager/package.json
konekoya: It'd be great if we can watch both :)
It's actually not a big deal for now as we don't change ohara-manager itself too often. But if we need to work on ohara-manager someday without the ability to watching the file changes and reload, we will need to manually kill it and restart whenever a change is made...

ohara-manager/package.json
saivirtue: Weird...it seems nodejs couldn't find the right path for us. I will add an issue to trace this in #4342

ohara-manager/package.json
saivirtue: Got it :)

Issue #4431, opened 2020-03-24
[GRADLE] test task should NOT depend on source jar and doc jar


chia7712:
source jar and doc jar are useless in testing so we should skip to build them.

1. - [x] for 0.10.0 (#4432)
1. - [x] for 0.9.1 (#4433)

chia7712:
I noted this issue when tracing the lock issue caused by ```gradle docJar```. By removing useless dependencies, the lock issue is resolved also. Not sure the root cause but I have no time to trace it :(

PR #4433, opened 2020-03-24
#4431 [GRADLE] test task should NOT depend on source jar and doc jar

chia7712:
backport to 0.9.1

related to #4431


chia7712:
retry

PR #4536, opened 2020-04-14
#4046 Enable shabondi metrics

vitojeng:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #4046)


ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/sink/DataGroup.scala
```diff
@@ -21,12 +21,29 @@ import java.util.concurrent.atomic.AtomicBoolean
...
+  objectKey: ObjectKey,
+  brokerProps: String,
+  topicNames: Seq[String],
+  pollTimeout: JDuration
+) extends Releasable {
+  private val log = Logger(classOf[RowQueue])
+
+  private val rowCounter: Counter =
+    Counter.builder
+      .key(objectKey)
```
vitojeng, line 38: @chia7712 Does it is OK to use Shabondi's objectKey in the Counter?

chia7712: sure.

BTW, the purpose of object key here is used to "distinguish" ohara metrics from other metrics if the jvm runs other process also (for example, worker and connector). the scenario of Shabondi get simpler here but it is fire to use ```Connector```.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/sink/DataGroup.scala
```diff
@@ -44,6 +61,7 @@ private[sink] class DataGroup(val name: String, brokerProps: String, topicNames:
   override def close(): Unit =
     if (closed.compareAndSet(false, true)) {
       producer.close()
```
chia7712, line 63: it would be better to swallow all exceptions temporarily before closing all resources. Those exceptions can squashed to single one before throwing.

vitojeng: Pardon me, not sure I'm catch your point. Can you give a example ?

vitojeng:
> Does route can access metrics now? Or it is addressed by another PR?

Not include in this PR. Will address in following up PR.

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/sink/DataGroup.scala
chia7712:
```scala
Releasable.close(producer)
Releasable.close(rowCounter)
```
this makes sure all resources are closed even if exception happens. For another, the thrown exceptions can be amassed and then merged to one exception. For example:
```scala
      var e: Throwable = null
      Releasable.close(producer, (exception: Throwable) => if (e == null) e = exception else e.addSuppressed(exception))
      if (e != null) throw e
```

the above code may be not necessary if you feel it is fine to just swallow exception in closing objects.

vitojeng: Got it, thanks @chia7712 !

ohara-shabondi/src/main/scala/oharastream/ohara/shabondi/sink/DataGroup.scala
```diff
@@ -41,9 +59,16 @@ private[sink] class DataGroup(val name: String, brokerProps: String, topicNames:
...
+  override def close(): Unit = {
     if (closed.compareAndSet(false, true)) {
-      producer.close()
+      var exception: Throwable = null
+      val addSuppressedException: Consumer[Throwable] = (ex: Throwable) => {
+        if (exception == null) exception = ex else exception.addSuppressed(ex)
+      }
+      Releasable.close(producer, addSuppressedException)
+      Releasable.close(rowCounter, addSuppressedException)
+      if (exception!=null) throw exception;
```
vitojeng, line 70: @chia7712 Please review again, thanks!

vitojeng: Still have some problem, will fix again later.

chia7712, line 70: ```exception!=null``` 

this may fails due to style check :(

chia7712: ```exception != null```

vitojeng: Ya, I just found it too.

vitojeng: @chia7712 Thanks for the review!

vitojeng:
not related error

Issue #4508, opened 2020-04-09
[Manager] Writing tests for redux epic


saivirtue:
It's important to make sure our epics logic could be run as expected:
- [x] writing tests for appEpics (#4517)
- [x] writing tests for brokerEpics (#4537)
- [x] writing tests for devToolEpics (#4551)
- [x] writing tests for connectorEpics (#4649)
- [x] writing tests for eventLogEpics (#4542)
- [x] writing tests for fileEpics (#4601)
- [x] writing tests for nodeEpics (#4595)
- [x] writing tests for streamEpics (#4666)
- [x] writing tests for pipelineEpics (#4704)
- [x] writing tests for snackbarEpics (#4591)
- [x] writing tests for topicEpics (#4655)
- [x] writing tests for workerEpics (#4544)
- [x] writing tests for workspaceEpics (#4577)
- [x] writing tests for zookeeperEpics (#4543)
- [x] re-enable deleteWorkspaceEpic (#4679)
- [x] trigger event log for all epics when throwing exception (#4698)

PR #4544, opened 2020-04-15
#4508 writing tests for workerEpics

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #4508)


ohara-manager/client/src/store/epics/worker/fetchWorkerEpic.js
```diff
@@ -57,6 +57,6 @@ export default action$ =>
   action$.pipe(
     ofType(actions.fetchWorker.TRIGGER),
     map(action => action.payload),
-    debounce(() => interval(1000)),
+    throttleTime(1000),
```
konekoya, line 60: Should we use debounceTime instead of throttleTime? As far as know they're different. But I might be wrong since i'm not very familiar with this piece of code

saivirtue: 就我的想法，`fetchWorker`的動作在時間間隔內如果被觸發多次，我們應該忽略在這段期間內任何的其他actions (因為get list理論上都應該是全部的資料，不用區別先後request的哪一個要真正執行)。
如果這邊用`debounceTime`，會讓fetch list的動作一定要等1秒後才會執行(因為這個function會捨棄其他的action)，並且只會執行最後一個；我想throttleTime應該就可以了 :)

konekoya: Right, after some googling, seems like `throttleTime` is a better choice as compare to `debounceTime` since it emits right away without that extra waiting. Good call 👍

Issue #4699, opened 2020-04-30
change the return type of TaskSetting#topicNames from List to Set


chia7712:
it makes sure the duplicate name disappear :)

1. - [x] for 0.11.0 (#4754)
1. - [x] for 0.10.0 (#4778)

PR #4778, opened 2020-05-10
#4699 change the return type of TaskSetting#topicNames from List to S…

chia7712:
backport #4699 to 0.10.0


chia7712:
the flaky is traced by #4784

@eechih It is ok to merge :)

Issue #4786, opened 2020-05-11
Apply Reference.NODE to all kind services


vitojeng:
In #4736, we decide add a new reference type `NODE` and apply into Shabondi's _NODE_NAMES_DEFINITION_.

For consistence reason, we should apply the `NODE` reference to _NODE_NAMES_DEFINITION_(key is **nodeNames**) in following services
* zookeeper
* broker 
* worker 
* stream

chia7712:
this is blocked by #4852

vitojeng:
@eechih 
Just update description, please take a look.
Will submit after #4852 is done.

vitojeng:
@chia7712 
I just noticed when creating or updating objects we use both **objectChecker.checkList#nodeNames()** and **objectChecker.checkList#references()**.

For example, in ShabondiRoute:

```scala
    objectChecker.checkList
      .nodeNames(creation.nodeNames)
      .brokerCluster(creation.brokerClusterKey)
      .references(creation.settings, creation.definitions)
      .check()
```

Does that means we perform the node validation twice? 
Just want to make sure I'm correct. That should not a big problem, IMO.

chia7712:
> Just want to make sure I'm correct.
That should not a big problem, IMO. We can improve in the future.

當該key存在的時候，的確會執行兩次，這個要等到我們允許在creation階段檢查所有參數的時候再來調整，也就是只需要呼叫reference check就好了

PR #4686, opened 2020-04-29
#4670 Update the Shabondi source and sink path segment

vitojeng:
### Checklist
- [x] [Review Contribution Guidelines](https://ohara.readthedocs.io/en/latest/contributing.html)
- [x] Add Reviewers (see [Authors](https://github.com/oharastream/ohara#ohara-team))
- [x] Add Assignees (of course, you are the assignee by default!)
- [x] DON'T set a Milestone
- [x] DON'T add any labels (PR labels will be added automatically by Jenkins)
- [x] Link to an issue(You can do so by using the required by keyword: required by #4670)
- [x] Add unit tests (or please explain why this PR is not worth having new tests)
- [ ] Pass QA (Sometimes you may encounter an existing flaky test which would obstruct you from getting QA passed. If the failed tests are unrelated, you can add a comment in the PR thread)

PR #4890, opened 2020-05-15
#4786 Apply Reference.NODE to all kind services

vitojeng:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #4686)


ohara-client/src/main/scala/oharastream/ohara/client/configurator/v0/package.scala
```diff
@@ -383,7 +384,8 @@ package object v0 {
   private[v0] def nodeDefinition: SettingDef.Builder => SettingDef =
     _.key(NODE_NAMES_KEY)
       .documentation("the nodes hosting this cluster")
-      .blacklist(Set(START_COMMAND, STOP_COMMAND, PAUSE_COMMAND, RESUME_COMMAND).asJava)
+      .optional(Type.ARRAY)
```
chia7712, line 387: the method ```blacklist``` define the type to ```array``` already. Hence, you don't need to call ```optional(Type.Array)``` again.

vitojeng: @chia7712 Got it.
Why we need a backlist here ?

chia7712: Some words, like "start" and "stop", are reserved by server so it always causes error. In such case, UI can offer friendly hint to users to avoid such behavior.

vitojeng: @chia7712 Thanks for explaination.
Does Shabondi's NODE_NAME_DEFINITION also need to add these blacklist?

chia7712: It would be better to set the black list for all node-related definition :)

vitojeng:
retry collieIT

vitojeng:
retry otherI

vitojeng:
retry otherIT

ohara-client/src/main/scala/oharastream/ohara/client/configurator/v0/package.scala
vitojeng: Got it.

vitojeng:
not related error

Issue #482, opened 2019-03-18
Enable to configure the timeout of ssh connection


vitojeng:
Currently, agent not set login session timeout value. There will be use default timeout value: 
```java
// org.apache.sshd.client.simple.SimpleClientConfigurator
long DEFAULT_AUTHENTICATION_TIMEOUT = Long.MAX_VALUE;   // virtually infinite
```

We need to set session timeout to prevent **infinite** timeout and make the timeout value configurable.

1. - [ ] for 0.11.0 (#4897)
1. - [ ] for 0.10.0 (#4915)

chia7712:
any update?

vitojeng:
Sorry, not yet. Hope to finish at v0.9.0

saivirtue:
Will this issue merge to `0.9` before today noon?
If not, please help me to move this issue to `0.10` :)

vitojeng:
@saivirtue Thanks for remind!

chia7712:
@vitojeng Let me address this for 0.10.0 as it does cause trouble to us :)

vitojeng:
> @vitojeng Let me address this for 0.10.0 as it does cause trouble to us :)

Thanks for great support!!

PR #4915, opened 2020-05-18
#482 Enable to configure the timeout of ssh connection (#4897)

chia7712:
backport #482 to 0.10.0


chia7712:
close as we are removing 0.10.x branch

Issue #4895, opened 2020-05-17
The default values are NOT added to response of Stream APIs


chia7712:
The default values from definitions should be added to response of Stream APIs

1. - [x] for 0.11.0 (#4896)
1. - [ ] for 0.10.0 (#4925)

PR #4925, opened 2020-05-18
#4895 The default values are NOT added to response of Stream APIs (#4…

chia7712:
backport #4895 to 0.10.0


chia7712:
close as we are removing 0.10.x branch

Issue #5131, opened 2020-05-31
Remove the override arguments from docker-compose build


chia7712:
Those arguments overrides nothing.

1. - [x] for 0.11.0 (#5132)
1. - [ ] for 0.10.0

PR #5133, opened 2020-05-31
#5131 Remove the override arguments from docker-compose build (#5132)

chia7712:
backport #5131 to 0.10.x0


chia7712:
@eechih 可否幫我合併？謝謝

Issue #5265, opened 2020-06-18
[Manager] Delete node should wait for node actually removed before close dialog


saivirtue:
1. Add node in node list of AppBar
2. remove the node which just added
2. the removing dialog will close after click button but the node is not really remove yet

Issue #5266, opened 2020-06-18
[Manager] Remove node should be disabled if there was service on it


saivirtue:
![image](https://user-images.githubusercontent.com/8088616/84999447-babcf800-b183-11ea-8b73-8961c856e8dc.png)

- Also add a test case to cover the logic of "node deletion loading" in #5265 (see [here](https://github.com/oharastream/ohara/pull/5283#discussion_r443300323))

saivirtue:
ping @chia7712 
When there are already some nodes existed (3 above), the getAll may contain the deletion object after deleting a node

chia7712:
> When there are already some nodes existed (3 above), the getAll may contain the deletion object after deleting a node

chia7712:
@saivirtue Did you wait for the response of deletion?

saivirtue:
> Did you wait for the response of deletion?

Hmm... We send the **delete** and **getAll** API simultaneously. Does it cause the getAll request fetched deletion objects?

chia7712:
> We send the delete and getAll API simultaneously.

maybe. They are executed at same time so getAll API may be executed before deleting object.

saivirtue:
> > We send the delete and getAll API simultaneously.
> 
> maybe. They are executed at same time so getAll API may be executed before deleting object.

If so, I thought it's the frontend bug, not backend. Do you plan to reduce the "possibility" of this failure in #5290?

vitojeng:
QA passed, thanks @saivirtue !

PR #5283, opened 2020-06-22
#5265 [Manager] Delete node should wait for node actually removed before close dialog

saivirtue:
### Checklist
- [x] DON'T set a Milestone
- [x] Link to an issue(You can do so by using the required by keyword: required by #5265)


ohara-manager/client/src/components/Node/NodeRemoveDialog.js
```diff
@@ -1,64 +0,0 @@
-/*
```
saivirtue, line 1: This file is duplicated with `NodeDeleteDialog`

ohara-manager/client/cypress/it/appBar/nodeList.test.ts
```diff
@@ -82,5 +82,39 @@ describe('App Bar', () => {
...
+        user: generate.userName(),
+        password: generate.password(),
+      };
+
+      // create workspace
+      cy.createWorkspace({ node });
+
+      // click node list
+      cy.findByTitle(/node list/i).click();
+      cy.findByTestId(`delete-node-${node.hostname}`).click();
```
saivirtue, line 99: This test will be changed in #5266, which will disable **node deletion which is in use**.
I will add another test for the "loading" case in #5266.
Anyway, it is OK to use this test case to make sure our loading logic is shown as expected.

saivirtue:
retry manager-it

saivirtue:
retry manager-e2e

Issue #5370, opened 2020-07-06
[Manager] Support build testing jars for copyJar command


saivirtue:
It's handier if we could combine `gradle ohara-it:jar` and `copy:jar`

PR #5371, opened 2020-07-06
#5370 [Manager] Support build testing jars for copyJar command

saivirtue:
### Checklist

- [x] [Review Contribution Guidelines](https://oharastream.github.io/en/docs/master/contributing/)
- [x] Add at least one reviewer (see [Authors](https://github.com/oharastream/ohara#ohara-team))
- [x] Add an assignee (of course, you are the assignee by default)
- [x] Do not set milestone, projects and labels (PR labels are added automatically by Jenkins)
- [x] Link to an issue
Only change shell script
- [x] Pass QA (If the failed tests are unrelated to this PR, add a comment in the PR thread)


ohara-manager/scripts/copyJars.js
```diff
@@ -25,14 +26,29 @@ const files = [
...
-
-  files.forEach((file) => {
-    fs.copyFileSync(`${srcBase}/${file}`, `${distBase}/${file}`);
+  const buildJars = execa('./gradlew', ['ohara-it:jar'], {
+    cwd: '..',
+    stdio: 'inherit',
   });
+
+  buildJars
+    .then((res) => {
```
konekoya, line 37: I'm not sure how to name this. But res (response) is probably not a good fit here...

ohara-manager/scripts/copyJars.js
```diff
@@ -25,14 +26,29 @@ const files = [
...
+        fs.copyFileSync(`${srcBase}/${file}`, `${distBase}/${file}`);
+      });
+    })
+    .catch((err) => {
+      console.log(err.message);
+      process.exit(1);
+    });
 };
 
 copyJars();
```
konekoya, line 54: Sam, could we refactor this and so it's just a script instead of a function? And so we won't need to call it later in the file :)

saivirtue: Sure thing :)
Will update later.

ohara-manager/scripts/copyJars.js
saivirtue: Maybe follow the definition of `then()`, using `value`? 🤔

konekoya: That will do :)

PR #5751, opened 2020-11-03
#5748 Automatically deploy volume to new node which is added by Broke…

### LINKED ISSUE #5748: Automatically deploy volume to new node which is added by Broker incremental API
If the brokers are mounted with existent volumes, the increment API (of brokers) will cause error since the volume is not deployed on the new node.

chia7712:
fix #5748


jackyoh (approved):

jackyoh:
retry collieIT

chia7712:
retry streamIT

PR #5781, opened 2020-11-13
#5780 Make maven central have higher priority

### LINKED ISSUE #5780: Make maven central have higher priority
Normally, we use local repo when testing snapshot version. Hence, the order of local repo should be lower than central repo. Otherwise, it may encounter error when local repo does not have specify jars

chia7712:
fix #5780


chia7712:
retry

jackyoh (approved):

PR #5786, opened 2020-11-16
#5782 [Manager] Cluster states are shown as 'unknown' in node's detail view

### LINKED ISSUE #5782: [Manager] Cluster states are shown as 'unknown' in node's detail view
### Current behavior
After a workspace is created, the node detail view is not showing all cluster state (worker, broker, zookeeper) and instead, only the worker state is displayed 

### Desired behavior 
All cluster state should correctly display in the node's detail view

### Steps to repro
1. Create a workspace
2. Open the node list page and select the node that your workspace is currently running at

I've noticed the behavior is causing by not calling `/brokers` and `/zookeepers` API when the app is initially loaded. If we try to open the workspace settings and then go back to the node list, the state will then correctly loaded and displayed in the view

Imgur

eechih:
fix #5782


eechih:
retry manager-it

konekoya (approved):
Nice work :)

eechih:
retry manager-it

